PR #2, opened 2024-02-27
Update protocol for backend

L-jasmine:
* I replaced std::sync::mpsc with crossbeam, as I believe it offers better performance.
* I added some new commands: Chat, and LoadModel.
* I moved the Sender for sending responses to the Command, which may make it clearer which command the response is coming from.


jmbejar:
> This PR is only for reference purposes for [PR1](https://github.com/project-robius/moxin/pull/1), and there is no need to merge it.

Gotcha! I'm currently looking at the comments and thinking in your suggestion number #3, which is an interesting one for sure. Will follow up in the original PR whenever makes sense, otherwise I will be commenting here.

PR #5, opened 2024-03-17
Implement some backend functions

L-jasmine:
I have implemented some basic chat functions to provide partial functionality testing. The feature for downloading remote model files is still under development.


moxin-protocol/src/open_ai.rs
```diff
@@ -40,7 +40,17 @@ pub struct ChatRequestData {
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum Role {
+    #[serde(rename = "system")]
+    System,
+    #[serde(rename = "user")]
+    User,
+    #[serde(rename = "assistant")]
+    Assistant,
 }
```
L-jasmine, line 54: I have changed the role in the `Message` & `MessageData` to an enum.

moxin-protocol/src/open_ai.rs
```diff
@@ -111,7 +121,7 @@ fn response_object() -> String {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ChunkChoiceData {
-    pub finish_reason: StopReason,
+    pub finish_reason: Option<StopReason>,
```
L-jasmine, line 124: I noticed in the OpenAI documentation that it might be a null, so I changed it to an Option here.

moxin-backend/src/backend_impls.rs
```diff
@@ -0,0 +1,716 @@
...
+            }
+            Command::StopLocalServer(tx) => {
+                Self::Model(ModelManagementCommand::StopLocalServer(tx))
+            }
+        }
+    }
+}
+
+#[test]
+fn test_chat() {
```
L-jasmine, line 533: This test demonstrates the basic usage, and if I have any misunderstandings regarding the parameters, they can be pointed out here.

moxin-backend/src/backend_impls.rs
```diff
@@ -0,0 +1,716 @@
...
+                                },
+                                like_count: 0,
+                                download_count: 0,
+                            },
+                            downloaded_at: chrono::Utc::now().date_naive(),
+                            compatibility_guess:
+                                moxin_protocol::data::CompatibilityGuess::PossiblySupported,
+                            information: String::new(),
+                        })
+                        .collect();
```
L-jasmine, line 672: Because our file server is not ready yet, most of the metadata here is fake.

L-jasmine:
The download link for the model I used is available here.

https://huggingface.co/second-state/Llama-2-7B-Chat-GGUF

L-jasmine:
@jmbejar Please review this pull request.

jmbejar:
> @jmbejar Please review this pull request.

@L-jasmine I'm kind of starting to review now. I took time to integrate with some progress for the frontend of the chat functionality and everything works OK 👏 👏 👏  Just wanted to confirm the thing runs well on my end.

I had to install things mentioned [here](https://wasmedge.org/docs/start/install). Maybe it is a good idea to drop some install notes in the Readme, but we can do it as part of this PR or later.

moxin-backend/src/backend_impls.rs
```diff
@@ -0,0 +1,716 @@
...
+                                model: String::new(),
+                                system_fingerprint: String::new(),
+                                object: "chat.completion.chunk".to_string(),
+                            })));
+                }
+                (Err(token_error), chat_completion_message) => {
+                    if let Some(chat_completion_message) = chat_completion_message.take() {
+                        let _ = self.token_tx.as_mut().unwrap().send(Ok(
+                            ChatResponse::ChatFinalResponseData(ChatResponseData {
+                                id: String::new(),
```
jmbejar, line 151: Should we autogenerate an id for the chat interaction? Maybe it can be done a bit later just to be fully compatible with the OpenAI format

moxin-backend/src/backend_impls.rs
```diff
@@ -0,0 +1,716 @@
...
+            }
+            Ok(n)
+        }
+
+        fn send_output(&mut self, output: Result<&[u8], TokenError>) -> bool {
+            if self.token_tx.is_none() {
+                return false;
+            }
+
+            match (output, &mut self.chat_completion_message) {
```
jmbejar, line 121: Nit: It took me a while to understand that the presence of `self.chat_completion_message` means the request was made with `stream: false`. An idea to improve readability could be to organize this function like:

```rs
fn send_output(&mut self, output: Result<&[u8], TokenError>) -> bool {
    if self.token_tx.is_none() {
        return false;
    }
            
    if chat_completion_message.is_some() {
        self.send_completion_output(output)
    } else {
        self.send_streamed_output(output)
    }
}
```

This would lead to those more scoped functions, `send_completion_output` and `send_streamed_output`, each one pattern matching only in the `token`.

jmbejar (approved):
@L-jasmine Good job with the PR. I haven't noticed this also implemented the streamed version of the chat, so I'm testing it now. So far so good 👍 

I have left comments, but nothing blocking the PR to be merged in my opinion. So, I'm approving. 

Go through the comments and make amends if you want, but please merge it when you're ready.

moxin-backend/src/backend_impls.rs
```diff
@@ -0,0 +1,716 @@
...
+                        self.token_tx
+                            .as_mut()
+                            .unwrap()
+                            .send(Ok(ChatResponse::ChatResponseChunk(ChatResponseChunkData {
+                                id: String::new(),
+                                choices: vec![ChunkChoiceData {
+                                    finish_reason: None,
+                                    index: 0,
+                                    delta: MessageData {
+                                        content: String::from_utf8_lossy(token).to_string(),
```
jmbejar, line 136: I noticed the first chunks comes with some blank spaces at the beginning. For instance, these are the two first chunks I'm receiving in one run:

```
        ChunkChoiceData {
            finish_reason: None,
            index: 0,
            delta: MessageData {
                content: " ",
                role: Assistant,
            },
            logprobs: None,
        },
        
        ChunkChoiceData {
            finish_reason: None,
            index: 0,
            delta: MessageData {
                content: " Gra",
                role: Assistant,
            },
            logprobs: None,
        },
```

Do you know the reason?

In the meanwhile, I will trim the answer in the UI to make sure we don't show extra whitespaces.

moxin-backend/src/backend_impls.rs
L-jasmine: I don't understand the generation format of the ID field in OpenAI, can I use a random UUID instead?

moxin-backend/src/backend_impls.rs
L-jasmine: Actually, I'm not sure what the reason is either; in some cases, the model's output does start with a space token. I think perhaps a solution would be to ignore these space tokens when displaying them until the first non-space token appears

moxin-backend/src/backend_impls.rs
jmbejar: Using random UUID seems OK to me 👍

moxin-backend/src/backend_impls.rs
jmbejar: Sure, no problem at all. Thanks for confirming it is an issue we should address in the frontend.

juntao (approved):
I will approve and merge this PR so that we can continue the work on a common branch. The remaining issues are minor and should be addressed in subsequent PRs. Thanks!

PR #10, opened 2024-04-01
Saved the downloaded file information to SQLite.

L-jasmine:
Saved the downloaded file information to SQLite. So now the downloadedFile data is all real.


L-jasmine:
@jmbejar There are some type changes in the protocol, and some parts of the frontend need to be synchronized. 
The reason I don't modify the frontend is because my VSCode will automatically execute `cargo fmt`, but the code in the frontend section doesn't seem to follow this style, so there will be many formatting changes, and the actual changes will become less apparent.

L-jasmine:
I found that my load model function has not been synchronized with the latest updates, so this PR cannot be merged yet.

moxin-backend/src/store/remote.rs
```diff
@@ -0,0 +1,278 @@
...
+        super::models::Model,
+        super::download_files::DownloadedFile,
+        Sender<anyhow::Result<FileDownloadResponse>>,
+    )>,
+) {
+    while let Ok((model, mut file, tx)) = rx.recv() {
+        let file_id = file.id.clone();
+        let mut send_progress = |progress| {
+            let _ = tx.send(Ok(FileDownloadResponse::Progress(
+                file_id.clone(),
```
jmbejar, line 205: There is a simple way to avoid passing the whole `Arc` here.

`(*file_id).clone()`

I wouldn't worry to much with having extra strings cloned for a while, because it is not much data involved.

If we do this, there is no need to add `Arc` into the protocol itself. I think we should leave frontend and backend decoupled in terms of concurrency aspects.

L-jasmine: I just felt that there were a lot of clones here, so I added an `Arc`.

moxin-protocol/src/protocol.rs
```diff
@@ -1,11 +1,12 @@
...
 use crate::data::*;
 use crate::open_ai::*;
+use anyhow::Result;
+use std::sync::mpsc::Sender;
+use std::sync::Arc;
 
 #[derive(Clone, Debug)]
 pub enum FileDownloadResponse {
-    Progress(FileID, f32),
+    Progress(Arc<FileID>, f32),
```
jmbejar, line 9: See my related comment with an idea to avoid adding the `Arc` here, so the protocol is free of shared state stuff like Arc, Mutex, etc

L-jasmine: Okay, I'll revert it in my next commit. Maybe there's no need to consider these issues for now.

L-jasmine: I have reverted it.

L-jasmine:
Now this PR is ready to be merged.

jmbejar (approved):
I was doing some basic testing, so I'm ready to merge.

PS I have found some issues loading models that are downloaded in subfolders, like we're doing in this PR but I will prepare a fox for that during the day

PR #13, opened 2024-04-04
Add more features to chat

jmbejar:
Among other things, we are adding the ability to delete and edit messages.

### DIFF
--- a/moxin-frontend/src/chat/chat_panel.rs
+++ b/moxin-frontend/src/chat/chat_panel.rs
@@ -16,40 +16,6 @@ live_design! {
 
     ICON_PROMPT = dep("crate://self/resources/icons/prompt.svg")
 
-    ChatLineBody = <View> {
-        width: Fill,
-        height: Fit,
-        spacing: 5,
-        flow: Down,
-
-        <View> {
-            height: 20,
-            align: {x: 0.0, y: 0.5},
-
-            role = <Label> {
-                width: Fit,
-                height: Fit,
-                draw_text:{
-                    text_style: <BOLD_FONT>{font_size: 10},
-                    color: #000
-                }
-            }
-        }
-
-        chat_label = <Label> {
-            width: Fill,
-            height: Fit,
-            padding: {top: 12, bottom: 12},
-
-            draw_text:{
-                text_style: <REGULAR_FONT>{font_size: 10},
-                color: #000,
-                word: Wrap,
-            }
-            text: "Chat Line"
-        }
-    }
-
     ChatAgentAvatar = <RoundedView> {
         width: 20,
         height: 20,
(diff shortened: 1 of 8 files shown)


moxin-frontend/src/chat/chat_panel.rs
```diff
@@ -254,8 +248,8 @@ pub struct ChatPanel {
     #[rust]
     auto_scroll_cancellable: bool,
 
-    #[rust(true)]
-    prompt_enabled: bool,
+    #[rust]
+    is_chat_streaming: bool,
```
fmzbl, line 252: (This may be out of the scope of the PR, we can discuss it later or smth, just a thought from a quick glance)

These 4 flags are starting to make the rest of the widget a bit messy and hard to follow, does it make sense to go with a sort of state machine approach here? maybe a full FSM is a bit of an overkill, but we could start encapsulating the state in an enum and have a couple of functions in the struct to handle stuff like the scrolling and state transitioning. Again, not wanting to make stuff overcomplicated here so lmk if im overthinking. 
```rust
enum ChatPanelState {
    Loading,
    Ready { auto_scroll_pending: bool, auto_scroll_cancellable: bool },
    Streaming,
}
```

fmzbl:
hey! I wasn't able to run this because it doesn't find `ViewAction::FingerHoverIn` on the makepad-widgets package in the moxin-changes branch. But I'll comment on what i saw from a brief look.

moxin-frontend/src/chat/chat_panel.rs
```diff
@@ -296,16 +290,30 @@ impl Widget for ChatPanel {
...
+                .chars()
+                .next()
+                .unwrap()
+                .to_uppercase()
+                .to_string();
         } else {
             chat_history = vec![];
             model_filename = "".to_string();
+            initial_letter = "".to_string();
         };
```
fmzbl, line 308: we could get rid of the mutable variables with the [map_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or) method, though im totally find with the imperative approach, its in most places in this code and i find it completely readable, so take this just as a curiosity.

```rust
let (chat_history, model_filename, initial_letter) =
    store
	.current_chat
	.as_ref()
	.map_or((vec![], "".to_string(), "".to_string()), |chat| {
	    let model_filename = chat.model_filename.clone();
	    let initial_letter = model_filename
		.chars()
		.next()
		.unwrap_or_default()
		.to_uppercase()
		.to_string();
	    (chat.messages.clone(), model_filename, initial_letter)
	});
```

fmzbl (commented):
looks good! just a thought about the ChatPanel, which i wouldn't necessarily address it in this pr.

fmzbl (approved):

joulei (approved):

jmbejar:
> hey! I wasn't able to run this because it doesn't find `ViewAction::FingerHoverIn` on the makepad-widgets package in the moxin-changes branch. I'll see if i can fix it, but I'll comment on what i saw from a brief look.

Cargo.lock changes should fix this

PR #16, opened 2024-04-05
Improve ChatPanel code using a state field

jmbejar:
Based on https://github.com/project-robius/moxin/pull/13#discussion_r1552732356


fmzbl (approved):
Looks great! feels more natural and robust without too much abstraction, i love enums.

PR #18, opened 2024-04-09
Add button to cancel response generation.

jmbejar:
This will be completelly done when backend support is available.

In addition, the send message button inside the prompt input is now clikeable and working.


L-jasmine:
I think this pull request can be merged first, and I will quickly support this feature on the backend.

PR #19, opened 2024-04-10
Support multiple file downloads and stop chat

L-jasmine:
The current parallel file download is a very simple version. If this part requires performance optimization in the future, I will optimize it to an async version based on Tokio.


L-jasmine:
@jmbejar Please review this pull request.

jmbejar (approved):
Tested it and works as expected, thanks!

I'm thinking now how we should approach the concurrent downloads limit in the user interface. Let's see if we want to do this or iterate this solution as you said

PR #23, opened 2024-04-12
[On hold] Better default downloads path

joulei:
Sets the default downloads path using $HOME/moxin/model_downloads, it currently uses the current directory.

Stores the path in the current Store so it can be used for display.

image

### DIFF
--- a/moxin-frontend/src/app.rs
+++ b/moxin-frontend/src/app.rs
@@ -109,19 +109,13 @@ live_design! {
 
 app_main!(App);
 
-#[derive(Live)]
+#[derive(Live, LiveHook)]
 pub struct App {
     #[live]
     ui: WidgetRef,
 
     #[rust]
     store: Store,
-}
-
-impl LiveHook for App {
-    fn after_new_from_doc(&mut self, _cx: &mut Cx) {
-        self.store = Store::new();
-    }
 }
 
 impl LiveRegister for App {
--- a/moxin-frontend/src/data/store.rs
+++ b/moxin-frontend/src/data/store.rs
@@ -1,11 +1,17 @@
 use crate::data::{chat::Chat, download::Download, search::Search};
+use anyhow::{Context, Result};
 use chrono::Utc;
 use makepad_widgets::DefaultNone;
 use moxin_backend::Backend;
 use moxin_protocol::data::{DownloadedFile, File, FileID, Model};
 use moxin_protocol::protocol::{Command, LoadModelOptions, LoadModelResponse};
 use std::collections::HashMap;
+use std::path::PathBuf;
 use std::sync::mpsc::channel;
+use std::{env, fs};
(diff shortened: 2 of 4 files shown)


moxin-frontend/src/app.rs
```diff
@@ -118,12 +118,6 @@ pub struct App {
     store: Store,
 }
 
-impl LiveHook for App {
-    fn after_new_from_doc(&mut self, _cx: &mut Cx) {
-        self.store = Store::new();
-    }
-}
-
```
joulei, line 126: Removing the setup from `after_new_from_doc` because one unnecessary instance of the Store (and therefore a backend) were being created by deriving `Default` and being called by Makepad on initialization of the App.

Another solution would be to make Store and `Option` in `App`, but that introduces conditionals **everywhere**.

Instead I moved the creation to the custom `Default` implementation.

moxin-frontend/src/data/store.rs
```diff
@@ -1,11 +1,17 @@
...
 use makepad_widgets::DefaultNone;
 use moxin_backend::Backend;
 use moxin_protocol::data::{DownloadedFile, File, FileID, Model};
 use moxin_protocol::protocol::{Command, LoadModelOptions, LoadModelResponse};
 use std::collections::HashMap;
+use std::path::PathBuf;
 use std::sync::mpsc::channel;
+use std::{env, fs};
+
+const DEFAULT_DOWNLOADS_FOLDER: &str = "moxin/model_downloads";
```
jmbejar, line 13: Don't we want to use a hidden folder instead, at least in MacOS. I see it would be more natural to have `.moxin` in the home folder. I don't remember what is the common practice in Win

joulei: Good point, I guess we don't want to have users fiddle with the files in the folder as it might leave the db state inconsistent. 
I updated the code using `.moxin` and using a hidden attribute for directories on Windows, I tried the Windows separately from Moxin, I can't get Moxin to run on Windows easily, it's mostly WasmEdge and linking errors.

Either way, if this breaks for Windows whenever someone tries it, we can go back and fix it.

joulei: Good point, I guess we don't want to have users fiddle with the files in the folder as it might leave the db state inconsistent. 

I updated the code using `.moxin`. For Windows, `.name` is a valid directory name but it won't make it hidden by default. To achieve changing visibility we'd need external crates that have APIs for modifying directory metadata. So I'm skipping that on Windows for now.

(I had pushed a commit earlier thinking I was able to do it without a crate, tried with and without a `winapi`, and got confused by the similar API names with `std`)

jmbejar (approved):

fmzbl (approved):

joulei:
Reopening in a new PR

PR #24, opened 2024-04-12
Fetch downloaded files after completion

joulei:
- Updates the Store with latest downloaded files, so that Chat and My Model screen update immediately

- Removes completed downloads from current downloads

### DIFF
--- a/moxin-frontend/src/data/download.rs
+++ b/moxin-frontend/src/data/download.rs
@@ -66,7 +66,7 @@ impl Download {
         });
     }
 
-    pub fn update_download_progress(&mut self) {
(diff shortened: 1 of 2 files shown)


moxin-frontend/src/data/download.rs
```diff
@@ -66,7 +66,7 @@ impl Download {
         });
     }
 
-    pub fn update_download_progress(&mut self) {
+    pub fn process_download_progress(&mut self) {
```
joulei, line 69: Minor (and very opinionated) renaming - I'd argue it's job is to process progress updates, and the backend is the one updating the progress

jmbejar (approved):

PR #26, opened 2024-04-12
Add Modal widget and view all modal

fmzbl:
First version of the modal, right now we are setting the model id by hand dispatching another action, this should be replaced in the future for a more direct approach, maybe passing data directly on the modal action.


moxin-frontend/src/app.rs
```diff
@@ -53,54 +57,68 @@ live_design! {
...
-                    height: Fill,
+                        width: Fill,
+                        height: Fill,
 
-                    discover_frame = <LandingScreen> {visible: true}
-                    chat_frame = <ChatScreen> {visible: false}
-                    my_models_frame = <MyModelsScreen> {visible: false}
+                        tab1_frame = <LandingScreen> {visible: true}
+                        tab2_frame = <ChatScreen> {visible: false}
+                        tab3_frame = <MyModelsView> {visible: false}
```
jmbejar, line 112: Looks like we're reverting some renaming done recently by @joulei

moxin-frontend/src/landing/model_card.rs
```diff
@@ -304,3 +387,81 @@ impl Widget for ModelCard {
...
+
+impl WidgetMatchEvent for ModelCard {
+    fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
+        let widget_uid = self.widget_uid();
+
+        if self.link_label(id!(view_all_button.link)).clicked(actions) {
+            cx.widget_action(
+                widget_uid,
+                &scope.path,
+                ModalAction::ShowModalView(id!(model_card_view_all_modal_view)[0]),
```
jmbejar, line 399: I'm pretty sure `id!(model_card_view_all_modal_view)[0]` is equivalent to `live_id!(model_card_view_all_modal_view)`, in case you want to avoid the `[0]` stuff

jmbejar (approved):
There is some renaming changes I would fix before merging, but it's OK in general 👏

moxin-frontend/src/app.rs
fmzbl: opps, i messed up the merge, fixing it.

joulei (approved):

PR #27, opened 2024-04-15
Dev/move file

L-jasmine:


L-jasmine:
@jmbejar Please review this pull request. thanks. If you have any suggestions for the MoveDownloadedFile, we can discuss them here.

jmbejar:
> @jmbejar Please review this pull request. thanks. If you have any suggestions for the MoveDownloadedFile, we can discuss them here.

Sorry for the delay but we will have some review from @joulei who was working on something similar on our end. It should happen today.

joulei:
Hi, to confirm the idea here: when changing download location, would the flow be:

1. Frontend initializes a second Backend instance with the new user-input models dir
2. Frontend calls MoveDownloadedFile on each DownloadedFile (on the new Backend instance), 
3. Backend handles each MoveDownloadedFile moving them to the current (new) models dir
4. Frontend swaps backend instances so that new downloads happen on the new models dir

Is that correct?
If the objective is to have the frontend "own" the downloads this makes sense.

With this approach I think will need the frontend to keep it's own db/config file with the current models dir. So that after the user changes it we can keep track of it for the next startup.
We might also need to store the previous models dir, for example if a user closes the app in the middle of moving files. We want to resume moving, and for that we need to read the previous directory.

moxin-backend/src/backend_impls.rs
```diff
@@ -795,15 +839,17 @@ pub struct BackendImpl {
...
     /// * `max_download_threads` - Maximum limit on simultaneous file downloads.
     pub fn build_command_sender(
+        home_dir: String,
         models_dir: String,
         max_download_threads: usize,
     ) -> Sender<Command> {
         wasmedge_sdk::plugin::PluginManager::load(None).unwrap();
 
-        let sql_conn = rusqlite::Connection::open(format!("{models_dir}/data.sql")).unwrap();
+        let sql_conn = rusqlite::Connection::open(format!("{home_dir}/data.sql")).unwrap();
```
joulei, line 852: Separating these is a good idea 👍🏽 

Side-note: since the data.sql is the actual database file, and not a set of SQL instructions, shouldn't the extension be something like `.sqlite`? (merely to avoid confusion)

L-jasmine: Okay, it seems like it's using .sqlite, which is a SQLite database.

L-jasmine:
> 2\. MoveDownloadedFile

The SQLite database I’m using contains the download paths for the files. Even if the models directory is changed, the files in the old models directory can still be loaded until they are moved. This means that multiple models directories can be effective simultaneously. Therefore, I provided a MoveDownloadedFile command. The act of moving files is not necessary for the operation of the program.

joulei (approved):
Sorry we let this go on for a couple of days, but it looks good to go 👍🏽, we just had to make sense of how we'd structure the FE around this feature.
We can merge after solving conflicts

jmbejar:
@L-jasmine I was thinking on resolving conflicts in this PR but most of them are caused by the change from `downloaded_path` to `download_dir`. Since you have a similar thing implemented in https://github.com/project-robius/moxin/pull/34, let's close this PR.

I will push to main some individual commits that are unrelated with the main purpose of the PR, but valuable ones (such us, filtering no-chat models).

The implementation of `MoveDownloadedFile` you can add it again in your new PR, or I can add it later based on this PR (so it won't be lost for sure).

Hope it makes sense and sorry for the delay with this PR

L-jasmine:
I will resubmit the content of this PR based on #34 later, because I feel that rewriting it will be faster than resolving the conflicts.

PR #30, opened 2024-04-17
Add external link widget. Add external link for user and model

fmzbl:


joulei (approved):

moxin-frontend/src/shared/external_link.rs
```diff
@@ -0,0 +1,104 @@
...
+        self.view.handle_event(cx, event, scope);
+        self.widget_match_event(cx, event, scope);
+    }
+
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        self.view.draw_walk(cx, scope, walk)
+    }
+}
+
+impl WidgetNode for ExternalLink {
```
jmbejar, line 65: This is the default implementation of `WidgetNode`, isn't it?
If we just use `#[derive(Live, LiveHook, Widget)]` in the `ExternalLink`... this block could be removed? I think so, but maybe I'm missing something

PR #34, opened 2024-04-21
Optimize the download manager

L-jasmine:


L-jasmine:
@jmbejar  This pull request has not been tested yet, and there might be some minor changes, so please do not merge it.

I submitted it because I made significant changes to your previous code. You can offer some suggestions, and we can discuss how to make better changes here.

L-jasmine:
I merged pending_downloads into downloads_files as per the TODO. This part of the functionality is not yet complete, mainly because get_all_pending_downloads does not retrieve files that are currently being downloaded.

moxin-backend/src/store/remote.rs
```diff
@@ -211,154 +209,189 @@ fn download_file(
...
+            let _ = token_tx.send(());
+        }
 
-            // Update our local database
-            {
-                let conn = sql_conn.lock().unwrap();
-                let pending_download = PendingDownloads {
-                    file_id: file_id.clone(),
-                    progress: progress,
-                    status: PendingDownloadsStatus::Downloading,
```
L-jasmine, line 273: This code is executed frequently during downloads, and it causes a large number of writes to the database and frequently occupies locks. So, I optimized this part of the logic. I store the total length of the file in the database, and when I need to get the download progress, I calculate it based on the downloaded portion and the total length.

jmbejar: Ok, I kind of have the same concern about frequent database access, although here is a situation where the database is only used by one client (unlike web servers database).

One import requirement to keep working is the ability to get this progress value (in percentage) when the whole app restarts. The user should see the paused downloaded with their corresponding progress. Agreed that it can be done by comparing the real file size and the stored total 👍

jmbejar:
> @jmbejar This pull request has not been tested yet, and there might be some minor changes, so please do not merge it.
> 
> I submitted it because I made significant changes to your previous code. You can offer some suggestions, and we can discuss how to make better changes here.

Thanks for the help again, and also thanks for sharing this partial progress... so we can track efforts.

I did a quick review and I like the direction, so please go ahead 👏

jmbejar:
@L-jasmine Any updates on this? I'm asking because we're looking to wrap up in the following days the implementation of the build we're going to demo in the NL rust conference. Please let me know if you need something from our end.

L-jasmine:
@jmbejar This pull request is ready to be merged.
If there are any bug, I will open a new pull request to fix them.

PR #35, opened 2024-04-23
Update My Models to latest designs

joulei:
- Updates the formatting of the files table
- Adds Delete Model modal (not yet implemented in the backend)

Does not include the info modal, I need to figure out the json formatting first.
I'll probably abstract a general modal prompt widget or rework the actions to include data, to avoid further duplication.


moxin-frontend/resources/icons/start_chat.svg
joulei, line 1: I couldn't get the SVG from figma to work for this one, using a placeholder.

PR #41, opened 2024-04-25
Implement review in finder

joulei:
Implements the open folder mechanism for navigating the model downloads.


fmzbl (approved):
LGTM. In the future we may have to consider other cases when its not windows nor macos, but its fine for now

PR #55, opened 2024-04-29
Implement resuming chat in my models table

joulei:
Adds a button to the My Models table, with a 'Resume Chat' action if there's a current active chat.

https://github.com/project-robius/moxin/assets/22042418/57989fa1-9db0-47fc-9071-83beff2a0167

### DIFF
--- a/moxin-frontend/src/data/store.rs
+++ b/moxin-frontend/src/data/store.rs
@@ -69,6 +69,8 @@ pub struct Store {
 
     pub preferences: Preferences,
     pub downloaded_files_dir: String,
+
+    pub active_chat_file: Option<FileID>,
 }
 
 impl Store {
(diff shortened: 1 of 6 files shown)


moxin-frontend/src/data/store.rs
```diff
@@ -69,6 +69,8 @@ pub struct Store {
 
     pub preferences: Preferences,
     pub downloaded_files_dir: String,
+
+    pub active_chat_file: Option<FileID>,
```
jmbejar, line 73: Not saying we should address it in this PR, but noticing we have `current_chat` representing the current chat session, but the information there is incomplete. We are storing the filename in this `Chat` instance but it could be a natural place to also store the `FileId`.

jmbejar (approved):
Noted a bit of redundancy in the design of `Store` but I wouldn't say it's a blocker. 👏 👏

moxin-frontend/src/data/store.rs
joulei: Good catch, I missed that.
I'll merge this as is, and rework it a bit on the delete model PR, I have to tinker around with those values so it'll be easier to test if I have this merged.

PR #56, opened 2024-04-29
Implement delete model file

joulei:
Implements the file deletion in the My Models table. 

Automatically refreshes state int he landing screen files.

If the model is actively loaded it will be ejected, and the chat panel will be reset.

https://github.com/project-robius/moxin/assets/22042418/53720c82-e8c7-4532-b0c4-e22011af8c17


jmbejar:
@L-jasmine I'm merging this change because we want to include it in an internal demo today, but please take a look at the addition to the backend and let us know if you would make it differently (although it was a very simple addition)

PR #63, opened 2024-05-02
Chat: show loading animation when waiting for first token

jmbejar:
Adds the loading animation when the chat model response is about to streaming:

image

(animation is working similar to Figma).


joulei (approved):

joulei:
Looks and works great! 👏🏽 
I think there's too little contrast between the colors and the background, but I wouldn't mind it until we have a newer design.

PR #68, opened 2024-05-03
Redesign chat screen when there are no downloaded models

fmzbl:
update to latest designs.

when there are downloaded models and none of them are selected:
- always show message input

Screenshot from 2024-05-03 05-48-08

when there are no downloaded models:
- update texts
-  always show message input
- show button "go to discover"
- update text color

image

### DIFF
--- a/moxin-frontend/src/chat/model_selector.rs
+++ b/moxin-frontend/src/chat/model_selector.rs
@@ -239,11 +239,29 @@ impl Widget for ModelSelector {
         let mut downloaded_files = store.downloaded_files.clone();
         downloaded_files.sort_by(|a, b| b.downloaded_at.cmp(&a.downloaded_at));
 
+        let choose_label = self.label(id!(choose.label));
         if downloaded_files.is_empty() {
-            self.label(id!(choose.label))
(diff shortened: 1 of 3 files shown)


moxin-frontend/src/chat/model_selector.rs
```diff
@@ -234,6 +234,36 @@ impl Widget for ModelSelector {
     }
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        let store = scope.data.get::<Store>().unwrap();
+
+        let mut downloaded_files = store.downloaded_files.clone();
+        downloaded_files.sort_by(|a, b| b.downloaded_at.cmp(&a.downloaded_at));
```
jmbejar, line 240: No need to sort here, right? 

Also, in this case it seems easy to avoid the clone, by doing:
`let no_available_models = store.downloaded_files.is_empty()`

fmzbl: oops, this is what happens when you recklessly copy paste, fixed

joulei (approved):

PR #81, opened 2024-05-20
Recover search backend errors

jmbejar:
This PR fixes issues with internet disconnection or unavailable server errors, which was breaking the Explore section once it happened.

Now, the user is informed when there is a connectivity issue. Also, the application keeps responsive to try again the search.


fmzbl (approved):
clean state management 👌

PR #82, opened 2024-05-23
Multiple chats

fmzbl:
- Update store to keep track of the chat history
- Refactor chat flow to support multiple chat history
- Create chat history widget
- Small fixes and style


moxin-frontend/src/data/chat.rs
```diff
@@ -24,29 +28,87 @@ impl ChatMessage {
...
+                    truncated.push_str(ellipsis);
+                    truncated
+                } else {
+                    message.content.clone()
+                };
+
+                self.set_title(title);
+            }
+        }
+    }
```
joulei, line 111: I think this is fine for now, we can iterate on this later and ask the model come up with a title based on the context.

moxin-frontend/src/data/store.rs
```diff
@@ -53,19 +55,21 @@ pub struct ModelWithPendingDownloads {
 
 #[derive(Default)]
 pub struct Store {
-    // This is the backend representation, including the sender and receiver ends of the channels to
-    // communicate with the backend thread.
+    /// This is the backend representation, including the sender and receiver ends of the channels to
+    /// communicate with the backend thread.
```
joulei, line 59: Thank you for catching these, I've been updating some as I find them

moxin-frontend/src/chat/chat_history.rs
```diff
@@ -0,0 +1,228 @@
...
+
+live_design! {
+    import makepad_widgets::base::*;
+    import makepad_widgets::theme_desktop_dark::*;
+
+    import crate::shared::styles::*;
+    import makepad_draw::shader::std::*;
+
+    ChatCard = {{ChatCard}} {
+        content = <RoundedView> {
```
joulei, line 15: Just a tip for the future, if the reason you're creating a `content = <RoundedView>` is that you want `ChatCard` to have the look/behaviour of `RoundedView`, you can do so directly by 'inhering':
`ChatCard = {{ChatCard}}<RoundedVIew> { ... }` in that way you don't have to deal with accessing `self.view(id!(content));`.
I wouldn't bother with changing this now though.

joulei (approved):
Great job! 
This looks good, thank you for cleaning up names and docs as you go 🫶

moxin-frontend/src/chat/chat_history.rs
fmzbl: thats exactly it yeah, thanks.

jmbejar (approved):
🚀 🚀 🚀

PR #83, opened 2024-05-23
Resume and Start chat buttons from landing page

noxware:
# Changes
- Replaces the "Downloaded..." button in the landing page with "Chat with Model" and "Resume".
  - These work the same as in the "My Models" screen.
- Tried to extract a common reusable button called `CButton` (to avoid name conflicts).
- Seems like a small issue with the cursor got fixed in the landing screen by using that button.

# Before
https://github.com/project-robius/moxin/assets/7684329/af64ed9d-1f7e-4e41-9c3b-1732e036581e

# After

https://github.com/project-robius/moxin/assets/7684329/fe195bac-465a-4275-96ab-145cf2b0ae63

https://github.com/project-robius/moxin/assets/7684329/ae417abe-3f57-406a-9f0c-eb0a1bc80b56

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2350,7 +2350,8 @@ dependencies = [
 [[package]]
 name = "wasmedge-macro"
 version = "0.6.1"
-source = "git+https://github.com/WasmEdge/wasmedge-rust-sdk.git?rev=1ef2434#1ef2434a84f1b78e769421ddf422a593593c6f68"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dbe80d95a88e9ac87b6aaf7bc9acd1fdfcd92045db2bf41a2262f623e2406a92"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2359,8 +2360,9 @@ dependencies = [
 
 [[package]]
 name = "wasmedge-sdk"
-version = "0.13.2"
-source = "git+https://github.com/WasmEdge/wasmedge-rust-sdk.git?rev=1ef2434#1ef2434a84f1b78e769421ddf422a593593c6f68"
+version = "0.13.5-newapi"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9a4a1bf8d775c0ff6aa2c15369ebf6c1dafe041be01aef1cc54734b395fceeed"
 dependencies = [
  "anyhow",
  "cfg-if",
@@ -2374,8 +2376,9 @@ dependencies = [
 
 [[package]]
 name = "wasmedge-sys"
-version = "0.17.5"
-source = "git+https://github.com/WasmEdge/wasmedge-rust-sdk.git?rev=1ef2434#1ef2434a84f1b78e769421ddf422a593593c6f68"
+version = "0.18.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1321e59863569b48630846734f3b77759e9f010e291c9ce3336fd77461307ce"
 dependencies = [
  "bindgen",
  "cfg-if",
@@ -2401,8 +2404,9 @@ dependencies = [
 
 [[package]]
 name = "wasmedge-types"
-version = "0.4.4"
-source = "git+https://github.com/WasmEdge/wasmedge-rust-sdk.git?rev=1ef2434#1ef2434a84f1b78e769421ddf422a593593c6f68"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c8bfc548253718d2013a8fa9cc9ccc00241e6216f36ee071fa8d7efe2b46c5f"
 dependencies = [
  "thiserror",
  "wat",
--- a/moxin-frontend/src/app.rs
+++ b/moxin-frontend/src/app.rs
@@ -159,13 +159,7 @@ impl LiveRegister for App {
         makepad_widgets::live_design(cx);
 
         // Shared
-        crate::shared::styles::live_design(cx);
-        crate::shared::resource_imports::live_design(cx);
-        crate::shared::widgets::live_design(cx);
-        crate::shared::icon::live_design(cx);
-        crate::shared::modal::live_design(cx);
-        crate::shared::external_link::live_design(cx);
-        crate::shared::popup::live_design(cx);
+        crate::shared::live_design(cx);
 
         // Landing
         crate::landing::shared::live_design(cx);
--- a/moxin-frontend/src/app.rs
+++ b/moxin-frontend/src/app.rs
@@ -247,6 +247,10 @@ impl MatchEvent for App {
                     self.store.download_file(file, model);
                     self.ui.redraw(cx);
                 }
+                ModelFileItemAction::Chat => {
+                    let chat_radio_button = self.ui.radio_button(id!(chat_tab));
+                    chat_radio_button.select(cx, &mut Scope::empty());
+                }
                 _ => {}
             }
 
--- a/moxin-frontend/src/app.rs
+++ b/moxin-frontend/src/app.rs
@@ -158,39 +158,10 @@ impl LiveRegister for App {
     fn live_register(cx: &mut Cx) {
         makepad_widgets::live_design(cx);
 
-        // Shared
         crate::shared::live_design(cx);
-
-        // Landing
-        crate::landing::shared::live_design(cx);
-        crate::landing::model_files_tags::live_design(cx);
-        crate::landing::model_files_item::live_design(cx);
-        crate::landing::model_files_list::live_design(cx);
-        crate::landing::model_files::live_design(cx);
-        crate::landing::model_card::live_design(cx);
-        crate::landing::model_list::live_design(cx);
-        crate::landing::landing_screen::live_design(cx);
-        crate::landing::search_bar::live_design(cx);
-        crate::landing::search_loading::live_design(cx);
-        crate::landing::sorting::live_design(cx);
-        crate::landing::downloads::live_design(cx);
-        crate::landing::download_item::live_design(cx);
-
-        // Chat
-        crate::chat::chat_screen::live_design(cx);
-        crate::chat::chat_panel::live_design(cx);
-        crate::chat::chat_line::live_design(cx);
-        crate::chat::chat_line_loading::live_design(cx);
-        crate::chat::chat_history::live_design(cx);
-        crate::chat::model_info::live_design(cx);
-        crate::chat::model_selector::live_design(cx);
-        crate::chat::model_selector_list::live_design(cx);
-
-        // My Models
-        crate::my_models::my_models_screen::live_design(cx);
-        crate::my_models::downloaded_files_table::live_design(cx);
-        crate::my_models::delete_model_modal::live_design(cx);
-        crate::my_models::model_info_modal::live_design(cx);
+        crate::landing::live_design(cx);
+        crate::chat::live_design(cx);
+        crate::my_models::live_design(cx);
     }
 }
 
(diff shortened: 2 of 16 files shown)


Cargo.lock
```diff
@@ -2350,7 +2350,8 @@ dependencies = [
 [[package]]
```
noxware, line 2350: My rust-analyzer insists on updating the lock file so maybe it's not up to date? Is this update correct?

moxin-frontend/src/app.rs
```diff
@@ -159,13 +159,7 @@ impl LiveRegister for App {
...
 
         // Shared
-        crate::shared::styles::live_design(cx);
-        crate::shared::resource_imports::live_design(cx);
-        crate::shared::widgets::live_design(cx);
-        crate::shared::icon::live_design(cx);
-        crate::shared::modal::live_design(cx);
-        crate::shared::external_link::live_design(cx);
-        crate::shared::popup::live_design(cx);
+        crate::shared::live_design(cx);
```
noxware, line 162: wdyt about doing the same for landing, etc and removing the comments used for grouping?

moxin-frontend/src/data/store.rs
```diff
@@ -9,7 +9,7 @@ use moxin_protocol::data::{
     DownloadedFile, File, FileID, Model, PendingDownload, PendingDownloadsStatus,
 };
 use moxin_protocol::protocol::{Command, LoadModelOptions, LoadModelResponse};
-use std::collections::{VecDeque, HashMap};
+use std::collections::{HashMap, VecDeque};
```
noxware, line 12: These kind of changes are from Rust's formatter. I think is nice to keep them.

moxin-frontend/src/my_models/downloaded_files_table.rs
```diff
@@ -271,6 +266,7 @@ impl Widget for DownloadedFilesTable {
     }
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        self.file_item_map.clear();
```
noxware, line 269: The hashmap I mentioned it could be leaking if it's not cleared each cycle :P

moxin-frontend/src/my_models/downloaded_files_table.rs
```diff
@@ -544,59 +540,45 @@ pub enum ButtonType {
 #[derive(Live, LiveHook, Widget)]
 pub struct ActionButton {
     #[deref]
-    view: View,
+    deref: CButton,
```
fmzbl, line 543: This is a NIT, but we are using the name of the widget we deref everywhere else, so it itches me a bit.

moxin-frontend/src/shared/mod.rs
```diff
@@ -6,3 +9,14 @@ pub mod resource_imports;
 pub mod styles;
 pub mod utils;
 pub mod widgets;
+
+pub fn live_design(cx: &mut Cx) {
```
fmzbl, line 13: I would discuss this kind of changes first, and then apply them for everything if we wanna do them.

moxin-frontend/src/shared/actions.rs
```diff
@@ -0,0 +1,9 @@
+use makepad_widgets::DefaultNone;
+use moxin_protocol::data::FileID;
+
+#[derive(Clone, DefaultNone, Debug)]
+pub enum DownloadedFileAction {
```
fmzbl, line 5: since this action is now being dispatched and defined outside of the original widget file, i think we should find a new name for it, it became something else.

moxin-frontend/src/shared/widgets/c_button.rs
```diff
@@ -0,0 +1,92 @@
...
+            icon_walk: {width: 12, height: 12}
+        }
+
+        label = <Label> {}
+    }
+);
+
+/// An attempt of a custom button, sharable with the rest of the app.
+#[derive(Widget, Live, LiveHook)]
+pub struct CButton {
```
fmzbl, line 37: Also, we are using views as buttons everywhere in the app, i think we should make this change in another pr, covering every case.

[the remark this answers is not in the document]
joulei: I'm also struggling to understand why CButton. 
Furthermore, I think I'd rather either keep a simple RoundedView (something simple quickly swappable for `Button`), or use Makepad's own `Button` already, I don't love that we're re-implementing a similar button over view, and adding similar but not consistent APIs (`tapped` vs `clicked`) is there something that prevents us from using `Button` today?

moxin-frontend/src/my_models/downloaded_files_table.rs
```diff
@@ -544,59 +540,45 @@ pub enum ButtonType {
...
 }
 
 impl ActionButtonRef {
     pub fn set_visible(&mut self, visible: bool) {
         if let Some(mut inner) = self.borrow_mut() {
-            inner.view.visible = visible;
+            inner.deref.visible = visible;
         }
     }
 }
```
joulei, line 580: Our current convention is to use `inner.view...`instead of `deref`, IMO it's a bit more explicit so that when navigating code I can quickly see over what we're operating and I don't need to go check separately what we're derefing. I think `deref` is more of a technicality than semantics.

moxin-frontend/src/shared/widgets/c_button.rs
```diff
@@ -0,0 +1,92 @@
+use makepad_widgets::*;
+
```
joulei, line 2: It took me a second to understand that `widgets.rs` has styling of existing widgets (no new widgets), and the new widgets folder has actual widgets.
I think we should either put new things like `c_button` in the root of `shared` module as we're already doing, or create a new submodule widgets (like you did now, and move all shared widgets there) but also rename `widgets.rs` or delete it and move its contents into styles.rs

moxin-frontend/src/shared/mod.rs
joulei: I think I brought this up before, it should be a good idea.
If we agree on the change, I wouldn't delete this now, but for next time I would do all of these in a separate PR.

moxin-frontend/src/my_models/downloaded_files_table.rs
```diff
@@ -544,59 +540,45 @@ pub enum ButtonType {
...
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
-        self.view.draw_walk(cx, scope, walk)
+        self.deref.draw_walk(cx, scope, walk)
+    }
+}
+
+impl WidgetMatchEvent for ActionButton {
+    fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
+        if self.deref.tapped(actions) {
```
joulei, line 561: Same as the other comment about self.`view` vs self.`deref`.

moxin-frontend/src/data/store.rs
```diff
@@ -49,6 +49,7 @@ pub struct DownloadInfo {
 pub struct ModelWithPendingDownloads {
     pub model: Model,
     pub pending_downloads: Vec<PendingDownload>,
+    pub current_file_id: Option<FileID>,
```
joulei, line 52: I might be missing something but can you use the already existing `current_chat` in `Store` instead of adding this to `ModelWithPendingDownloads`? Since in `draw_walk` of `ModelFilesList` you have access to the Store. It might  look like more work on the other side but semantically it doesn't seem like `ModelWithPendingDownloads` would need a `current_file_id`.

joulei (commented):
Appreciate you trying to find more ergonomic ways in Makepad.
However for this case I'd rather use Makepad's `Button` since it already has a `.clicked` API. I wouldn't be hesitant of having to create custom Widgets for lots of things if that means simplifying and shortening the code.
If the goal is to simplify and re-use, in this case it could be a lot simpler to make the `DownloadedFilesTable` rows their own Widget, and then just use `button.clicked()`. 

Don't necessarily take this a request of changes though, I'd be happy if we reach a middle ground were there's later a quick path to swap out these for Makepad's `Button` without much changes (see my comment about using RoundedView or matching API).

joulei (commented):
I know we need new design for this, but having different colors in `Download` and `Resume Chat` feels a bit off to me. Could we perhaps just either use the same color in the meantime, or just the outline like in `Chat with Model`? (so that it doesn't look like we made a typo/mistake with the colors)

Screenshot 2024-05-24 at 9 26 09 AM

moxin-frontend/src/app.rs
noxware: Taking that as a yes

moxin-frontend/src/data/store.rs
jmbejar: I'm almost sure there is no way to reach the `store` from `ModelFilesList`. This is because we're passing more granular data at this widget (represented with the struct `ModelWithPendingDownloads`). So, I think @noxware was following the existing approach there.

moxin-frontend/src/landing/model_files_item.rs
```diff
@@ -195,18 +215,33 @@ impl Widget for ModelFilesItem {
...
-
-                cx.widget_action(
-                    widget_uid,
-                    &scope.path,
-                    ModelFileItemAction::Download(file.clone(), model.clone()),
-                );
-            }
+        let widget_uid = self.widget_uid();
+
+        if self.cbutton(id!(download_button)).tapped(&actions) {
```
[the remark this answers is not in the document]
noxware: Yeah, I didn't want to imply a "mouse device" when this may be eventually be used in mobile so for me "tapped" or "[pressed](https://api.flutter.dev/flutter/material/ButtonStyleButton/onPressed.html)" are more "cross-platform" terms, but I agree that we should make everything consistent for now 👍

moxin-frontend/src/data/store.rs
jmbejar: Relevant code: https://github.com/project-robius/moxin/blob/main/moxin-frontend/src/landing/model_list.rs#L101-L103

(from this point deeper in the widgets tree, the scope does not hold the entire `store`)

moxin-frontend/src/landing/model_files_item.rs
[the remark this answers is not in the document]
noxware: Change done

moxin-frontend/src/shared/mod.rs
noxware: I was reading the comments in order and since I saw @joulei thumbs up I applied the change for the rest of occurrences but I didn't saw this comment.

Shall I revert https://github.com/project-robius/moxin/pull/83/commits/d52c1ecdddedef424e33c380250685d36df8e469?

moxin-frontend/src/data/store.rs
noxware: Yes, what @jmbejar mentioned is the reason why I choose to do it this way. The `draw_walk` and `handle_event` receive scopes with different things. The draw walk receives what @jmbejar mentioned.

I think the store is accesible in the handle event (which surprised me and confused me a lot until I followed the code up to discover the origin of both scopes), but adding a `#rust` field and setting it on events as a workaround was worst in my opinion.

Extra note:
I noticed that makepad, apart of having a `Scope.with_data` has a `Scope.with_props`. The difference is that one is mutable and the later is immutable. If they are hold in different slots, then, would this problem be resolved by using `props` for `ModelWithPendingDownloads` and reserve `data` to always pass the global context (store) down?

moxin-frontend/src/shared/actions.rs
joulei: Oh yeah I meant that the Action itself is the same one. (And it's technically an action on a downloaded file either way). But I do like `ChatAction` 👍🏽, it's a bit more obvious

moxin-frontend/src/shared/widgets/c_button.rs
noxware: Covered in slack discussion

joulei (approved):

PR #85, opened 2024-05-28
Refactor to make use of the original Makepad Button widget

jmbejar:
This PR replaces some of the buttons in the Explore and Download Manager sections to make use of the original widget from Makepad.

It defines a custom `MoxinButton` which is basically a `Button` widget with most of the shader code reimplemented to fit our application design. It follows what we did for the `TextInput` widget (where we have a `MoxinTextInput`).

Now, it has a better look&feel because hovering those buttons has animations in the colors. At the code level, it relying in most idiomatic Makepad code since it is using the `Button` API directly.

### DIFF
--- a/moxin-frontend/src/app.rs
+++ b/moxin-frontend/src/app.rs
@@ -64,21 +64,21 @@ live_design! {
 
                         discover_tab = <SidebarMenuButton> {
                             animator: {selected = {default: on}}
-                            label: "Discover",
+                            text: "Discover",
                             draw_icon: {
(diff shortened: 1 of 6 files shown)


moxin-frontend/src/app.rs
```diff
@@ -64,21 +64,21 @@ live_design! {
 
                         discover_tab = <SidebarMenuButton> {
                             animator: {selected = {default: on}}
-                            label: "Discover",
```
jmbejar, line 67: Changes in this file are simply a catch up with latest modifications in Makepad

PR #87, opened 2024-05-31
Progress and download controls for each model download item in the landing

noxware:
# Changes
- See download progress of each model file with a progress bar and text.
- Resume, pause, cancel and retry controls inline with the progress.

# Links
- [Figma](https://www.figma.com/design/LIaqW5HF3CGtq4XMpDxDMD/MoxinUI---LLM-Desktop-App?node-id=1332-17276&t=QvBkQt8twguLLYZh-0)

# Notes
- Although not in the design, the decimal precision has been added to the text progress because:
  - It's consistent with what we already have.
  - Without it, it's difficult to tell quickly if everything is working properly.
- Because of the previous point, the buttons may be placed a bit more on the right than in the design.

# Screenshots

image

### DIFF
--- a/moxin-frontend/src/landing/model_files_item.rs
+++ b/moxin-frontend/src/landing/model_files_item.rs
@@ -17,6 +17,11 @@ live_design! {
     START_CHAT = dep("crate://self/resources/icons/start_chat.svg")
     RESUME_CHAT = dep("crate://self/resources/icons/play_arrow.svg")
 
+    ICON_PAUSE = dep("crate://self/resources/icons/pause_download.svg")
+    ICON_CANCEL = dep("crate://self/resources/icons/cancel_download.svg")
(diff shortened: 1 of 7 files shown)


moxin-frontend/src/landing/model_files_item.rs
```diff
@@ -69,13 +74,73 @@ live_design! {
...
+
+            progress_fill = <RoundedView> {
+                width: 0,
+                height: Fill,
+                draw_bg: {
+                    radius: 2.5,
+                }
+            }
+        }
+        progress_text_layout = <View> {
```
noxware, line 112: Could not align the text properly to the right with empty pad in the left without using a View.

jmbejar (approved):
Nice! 🚀 🚀 🚀

PR #88, opened 2024-06-03
Refactor data/store.rs

jmbejar:
This PR has two phases of changes. In the first one:

* Logic to add fake logic extracted out of `Store` struct.
* Established the role of some of the fields of `Store`
  * `pending_downloads` is the source of truth for the Downloads Manager (files being downloaded)
  * `models` is the source of truth for the Explore section (search).
  * `downloaded_files` is the source of truth for My Models
* Move some draw logic from `ModelFilesList` to `ModelFilesItems` given it was all local logic at the items level.
* Extracted minor functions from `Store` and place them in the widgets where they were used, when appropriate. This was to reduce the size and responsibilities of `Store`

Then, I went ahead and extracted out most of the logic from `Store` to specific structs/modules:
- Preferences
- Downloads
- Search (augmenting the existing struct)
- Chats

Now, the `Store` is quite flat and short. Its responsibility is to hold the other objects and implement functions that coordinates or integrates between them.


moxin-frontend/src/landing/model_files_list.rs
```diff
@@ -228,3 +108,21 @@ impl ModelFilesListRef {
         inner.area.rect(cx).size.y
     }
 }
+
+fn model_featured_files(model: &ModelWithDownloadInfo) -> Vec<FileWithDownloadInfo> {
+    model
+        .files
+        .iter()
+        .filter(|f| f.file.featured)
+        .cloned()
```
noxware, line 117: Nitpick: Is clone necessary or can we just return something like

```rust
fn model_featured_files(model: &'m ModelWithDownloadInfo) -> impl Iterator<Item = &'m FileWithDownloadInfo> {
    model
        .files
        .iter()
        .filter(|f| f.file.featured)
}
```

moxin-frontend/src/data/preferences.rs
```diff
@@ -7,12 +7,14 @@ use std::{
...
 use serde::{Deserialize, Serialize};
 
-use super::filesystem::moxin_home_dir;
+use super::filesystem::{moxin_home_dir, setup_model_downloads_folder};
 const PREFERENCES_FILE: &str = "preferences.json";
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
 pub struct Preferences {
     pub current_chat_model: Option<FileID>,
+    #[serde(default = "setup_model_downloads_folder")]
```
noxware, line 16: Is it good to use a function with side effects as a serde default?
Also, this function is currently blocking but would this still work if it becomes non-blocking?

noxware (dismissed):
The PR is long but very clear and that deserves merit 👏

The overall organization looks good to me. As this is mainly a PR to improve the organization I will just leave some style nitpick comments and leave this approved.

moxin-frontend/src/data/preferences.rs
jmbejar: It's a good call. I'm reverting this

moxin-frontend/src/landing/model_files_list.rs
jmbejar: Confirmed that `cloned` (not `clone`) is need here to compile it.

I think it is related with this part of the `cloned` docs:

> Creates an iterator which [clone](https://doc.rust-lang.org/std/clone/trait.Clone.html#tymethod.clone)s all of its elements.
> This is useful when you have an iterator over &T, but you need an iterator over T.

noxware: > Confirmed that cloned (not clone)

Yeah, I was meaning if the action of cloning was necessary regardless of the name of the method implying the clone.

>  but you need an iterator over T.

Yeah, I mentioned this because I didn't notice a reason why this needs to be owned.

This is used just here right? https://github.com/moxin-org/moxin/blob/c1a8d2ef53eb859be1835c22a5f01a5cc1753691/moxin-frontend/src/landing/model_files_list.rs#L85

If you just need borrowed data you probably don't need ownership of the items. If it's being used just there, there shouldn't be lifetime issues. And if you are just accessing to the elements sequentially (`for i in 0..files_info.len()`) then anything implementing the `Iterator` trait should be able to be passes ad parameter.

(Of course what I mentioned will not compile if the other place is not changed).

But anyways, this is probably a less important comment that escaped my mind while reading the PR, and I didn't even tested this so please, proceed with merging this ignoring this less important comment.

PR #94, opened 2024-06-06
style - 05/17 chat design

noxware:
# Changes
- Adjust styles of the chat text input, message bubbles, and a bit the selector to match better the styles [here](https://www.figma.com/design/LIaqW5HF3CGtq4XMpDxDMD/MoxinUI---LLM-Desktop-App?node-id=1522-17301&t=SbVStHwoDV9PfV8f-4).
- The user avatar png is deleted as it is no longer used.

# Notes
- Making the text bubble fit the its label while not overflowing it's parent is currently not trivial so for now chat bubbles are full width with different margins on the left to give them different idents (instead of having one fit bubble on the left and another fit bubble on the right). 

# Screenshots
Screenshot 2024-06-06 at 18 06 30

Screenshot 2024-06-06 at 18 47 07


moxin-frontend/src/chat/chat_panel.rs
```diff
@@ -323,6 +366,9 @@ live_design! {
 
                 flow: Overlay
                 chat = <PortalList> {
+                    scroll_bar: {
+                        bar_size: 0.0,
```
noxware, line 370: The whole portal list showing messages has a scroll bar that was not noticeable when everything was white, but know that chat bubbles have borders and the user bubble is not white, it is very noticeable. This hides the scroll bar again.

src/chat/chat_panel.rs
```diff
@@ -323,6 +427,9 @@ live_design! {
 
                 flow: Overlay
                 chat = <PortalList> {
+                    scroll_bar: {
+                        bar_size: 0.0,
```
noxware, line 431: The whole portal list showing messages has a scroll bar that was not noticeable when everything was white, but know that chat bubbles have borders and the user bubble is not white, it is very noticeable. This hides the scroll bar again.

src/chat/chat_panel.rs
```diff
@@ -50,11 +51,67 @@ live_design! {
...
+                    visible: false,
+                }
+                bubble = {
+                    draw_bg: {
+                        color: #15859A
+                    }
+                    markdown_message_container = {
+                        markdown_message = {
+                            draw_normal: {
+                                color: (#fff),
```
noxware, line 71: I extracted colors out because each is specific to the kind of message and just overriding one of them looks like "patching"

src/chat/chat_panel.rs
```diff
@@ -50,11 +51,67 @@ live_design! {
     }
 
     UserChatLine = <ChatLine> {
+        margin: {left: 100}
```
noxware, line 54: Makes the message bubble of the user look more in the "right" by indenting it more than the assistant message bubble.

Ideally, the bubble should fit the text and be aligned to the right but it's not currently trivial to implement.

src/chat/chat_panel.rs
jmbejar: This makes sense 👍 

The Markdown widget is a complex one, so yeah.. I don't see a simpler way to adjust color at the moment.

jmbejar:
There is a bit of difference in the bubble's background color between this PR and Figma. One is white and the other gray:

Figma:
image

src/chat/chat_panel.rs
```diff
@@ -466,7 +573,6 @@ impl Widget for ChatPanel {
                         } else {
                             item = list.item(cx, item_id, live_id!(UserChatLine)).unwrap();
                             chat_line_item = item.as_chat_line();
-                            chat_line_item.set_role("You");
```
jmbejar, line 469: Changes in this PR would force us to thing a new name for the `set_role` function. Maybe `set_sender_name`?

jmbejar (approved):
Nice job!

PR #95, opened 2024-06-06
Make Moxin runnable from a macOS `.app` bundle.

kevinaboos:
~~**Note: do not merge this yet, it's messy.**~~ ready now

The problem is that when an app bundle is run, the current working directory is `/` (root), which is read-only on macOS. Thus, the attempt to access `./data.sqlite` is actually an attempt to create the file `/data.sqlite`, which immediately fails because a file cannot be created on the read-only root partition.

Instead of using the current working directory, this commit now changes file paths to be relative to `$HOME`, which will only really work properly on Unix-like environments where the `$HOME` environment variable is defined. Even on Unix-like systems, this isn't *guaranteed* to work, since various sandboxing or environment masking techniques may be in effect.

To actually create the bundle:
1. install cargo-bundle: `cargo install cargo-bundle`
2. run `cargo bundle` in the moxin repo directory.
3. patch the bundled moxin binary to depend directly on the installed wasmedge dylib:
    ```sh
    install_name_tool -change "@rpath/libwasmedge.0.dylib"  "$HOME/.wasmedge/lib/libwasmedge.0.dylib"  target/debug/bundle/osx/moxin.app/Contents/MacOS/moxin
    ```
    * Note that if you build in release mode with `cargo bundle --release`, the path will be `target/release/bundle...`.
    * This also assumes that you have installed wasmedge properly, which can be done via:
       ```sh
       curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install_v2.sh | bash
       source $HOME/.wasmedge/env
       ```
4. Now you can run the `moxin.app` bundle (located in `moxin/target/debug/bundle/osx/` by double-clicking it in Finder, or by running `open moxin.app`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ members = [
 name = "moxin"
 version = "0.1.0"
 edition = "2021"
-description = "A Client for Downloading and Chatting with AI LLMs"
+description = "A desktop GUI client for downloading and chatting with AI LLMs"
 
 [dependencies]
 moxin-protocol = { path = "moxin-protocol" }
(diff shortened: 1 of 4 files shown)


jmbejar (approved):
Thanks!!

PR #96, opened 2024-06-07
Refactor everything to use correct canonical OS directories

kevinaboos:


src/data/preferences.rs
```diff
@@ -71,7 +71,11 @@ fn write_to_file(json: &str) -> Result<(), std::io::Error> {
 }
 
 fn preferences_path() -> PathBuf {
-    let home_dir = moxin_home_dir();
-    let preferences_path = PathBuf::from(home_dir).join(PREFERENCES_FILE);
-    preferences_path
+    let preference_dir = project_dirs().preference_dir();
+    // TODO: this is overkill to create the directory every time.
+    //       This should be done somewhere during init, but I'm not sure where.
+    std::fs::create_dir_all(&preference_dir).unwrap_or_else(|_|
```
jmbejar, line 77: The actual folder creation (if needed) is done in a previous call to `setup_model_downloads_folder` (see https://github.com/moxin-org/moxin/pull/96/files#diff-9cac44800b86084f5e0e0a00ab63cc15615598154b89126aea415e93019b7ca0R25)

Not sure why you added this line, since it is not a responsibility of this function to do it.

jmbejar: I tested it because I didn't understand why it was not failing after being invoked more than once. I've just convinced to myself that `std::fs::create_dir_all` is not erroring when the folder already exists (although it is hard to confirm from the docs!).

So, we can remove this line later, but no real harm.

jmbejar (approved):

src/data/preferences.rs
kevinaboos: we can't remove this line, but it should be moved elsewhere. Previously when everything was in the user's `$HOME` directory, a single call covered it. But in this case, the preferences directory may be in a completely different location than the app's data directory, so we need at least one call to `fs::create_dir_all()`.

kevinaboos: btw, without this line here, i get a runtime panic on my mac.

kevinaboos: > I've just convinced to myself that `std::fs::create_dir_all` is not erroring when the folder already exists (although it is hard to confirm from the docs!).

yes that is correct, that's the behavior of `mkdir -p` on Unix-like OSes.

PR #97, opened 2024-06-11
Initial setup to build platform bundles

dvdhrm:
This adds a maintenance makefile that can build Linux Flatpaks with Moxin. I am working on similar targets for Windows Apps and macOS bundles.

Note that there are some difficulties involved:

1) Makepad hard-codes the absolute path to the source directory in its binaries. This is used to locate resources at runtime. This obviously will not work when distributing applications. Hence, this makefile collects resources and bundles them as part of the application. Currently, only resources of `moxin` and `makepad_widgets` are collected. This list is hard-coded, but can be extended easily.

2) We have to tell makepad about the location of the resources at runtime. This will require knowledge of the packaging format. Currently, this PR hard-codes the path so it works for Flatpaks, but this will certainly break other platforms / builds. We likely want feature-flags for each bundle we support, so our code can choose suitable locations.

3) We have to bundle wasmedge, since we cannot rely on it to be installed. The build currently downloads it from github and simply copies it into the flatpak. Not sure whether we can find better ways to do this.

This is a draft, but you can certainly test it right now. Simply checkout this branch and run `make flatpak-build` followed by `make flatpak-run`.


kevinaboos:
> (Added support to produce unsigned macOS bundles)

oh thanks, @dvdhrm i didn't realize you were still working on this, but i've been able to use [`cargo-packager`](https://github.com/crabnebula-dev/cargo-packager) to generate bundles and DMGs.

joulei:
Closing this PR since we now have a working packaging solution using `cargo-packager`

PR #98, opened 2024-06-11
fix - chat panel visibility states + refactor - local state reconciliation

noxware:
# Changes

- Centralized everything mutating visibility states of chat panel into a single method so it's easier to debug and to define what is shown in each possible state.
- Fixed multiple visibility overlap bugs.
- Model selector initially empty and visible.
- As agreed , state is kept in the widget as it was before giving the chance to change it from there, but is computed from store when an event happens (the mouse is moved, or whatever).
- Improved autoscroll behavior a bit. Autoscroll should now be reliable if you never scrolled up or if you did but pressed the jump to bottom button. Other issues are harder to fix without modifying the portal list widget.

### DIFF
--- a/src/chat/chat_panel.rs
+++ b/src/chat/chat_panel.rs
@@ -667,7 +667,7 @@ impl WidgetMatchEvent for ChatPanel {
             }
         }
 
-        self.jump_to_bottom_actions(cx, actions, scope);
+        self.jump_to_bottom_actions(cx, actions);
 
         match self.state {
             ChatPanelState::Idle => {
@@ -709,36 +709,11 @@ impl WidgetMatchEvent for ChatPanel {
 }
 
 impl ChatPanel {
-    fn jump_to_bottom_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
+    fn jump_to_bottom_actions(&mut self, cx: &mut Cx, actions: &Actions) {
         if let Some(fe) = self.view(id!(jump_to_bottom)).finger_up(actions) {
             if fe.was_tap() {
                 self.scroll_messages_to_bottom(cx);
                 self.redraw(cx);
-            }
-        }
-
-        let jump_to_bottom = self.view(id!(jump_to_bottom));
-        match self.state {
-            ChatPanelState::Streaming {
-                auto_scroll_pending: true,
-                ..
-            } => {
-                // We avoid to show this button when the list is auto-scrolling upon
-                // receiving a new message. Otherwise, the button flicks.
-                jump_to_bottom.set_visible(false);
-            }
-            ChatPanelState::Idle | ChatPanelState::Streaming { .. } => {
-                let store = scope.data.get_mut::<Store>().unwrap();
-                let has_messages = store
-                    .chats
-                    .get_current_chat()
-                    .map_or(false, |chat| !chat.borrow().messages.is_empty());
-
-                let list = self.portal_list(id!(chat));
-                jump_to_bottom.set_visible(has_messages && list.further_items_bellow_exist());
-            }
-            ChatPanelState::Unload => {
-                jump_to_bottom.set_visible(false);
             }
         }
     }
@@ -942,21 +917,30 @@ impl ChatPanel {
                 show!(no_downloaded_model);
                 hide!(no_model);
                 hide!(empty_conversation);
+                hide!(jump_to_bottom);
             }
             ComputedState::NoModelSelected => {
                 hide!(no_downloaded_model);
                 show!(no_model);
                 hide!(empty_conversation);
+                hide!(jump_to_bottom);
             }
             ComputedState::ModelSelectedWithEmptyChat => {
                 hide!(no_downloaded_model);
                 hide!(no_model);
                 show!(empty_conversation);
+                hide!(jump_to_bottom);
             }
             ComputedState::ModelSelectedWithChat => {
                 hide!(no_downloaded_model);
                 hide!(no_model);
                 hide!(empty_conversation);
+
+                if self.portal_list(id!(chat)).further_items_bellow_exist() {
+                    show!(jump_to_bottom);
+                } else {
+                    hide!(jump_to_bottom);
+                }
             }
         }
     }
--- a/src/chat/chat_panel.rs
+++ b/src/chat/chat_panel.rs
@@ -508,60 +508,9 @@ impl Widget for ChatPanel {
     }
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        self.switch_visibilities(cx, scope);
+
         let store = scope.data.get_mut::<Store>().unwrap();
-
-        #[derive(Debug)]
-        enum State {
-            NoModelsAvailable,
-            NoModelSelected,
-            ModelSelectedWithEmptyChat,
-            ModelSelectedWithChat,
-        }
-
-        let state = if store.downloads.downloaded_files.is_empty() {
-            State::NoModelsAvailable
-        } else if store.chats.loaded_model.is_none() {
-            State::NoModelSelected
-        } else {
-            store
-                .chats
-                .get_current_chat()
-                .map_or(State::ModelSelectedWithEmptyChat, |chat| {
-                    if chat.borrow().messages.is_empty() {
-                        State::ModelSelectedWithEmptyChat
-                    } else {
-                        State::ModelSelectedWithChat
-                    }
-                })
-        };
-
-        dbg!(&state);
-
-        match state {
-            State::NoModelsAvailable => {
-                self.view(id!(no_downloaded_model)).set_visible(true);
-                self.view(id!(no_model)).set_visible(false);
-                //self.view(id!(main)).set_visible(false);
-                self.view(id!(empty_conversation)).set_visible(false);
-            }
-            State::NoModelSelected => {
-                self.view(id!(no_downloaded_model)).set_visible(false);
-                self.view(id!(no_model)).set_visible(true);
-                //self.view(id!(main)).set_visible(false);
-                self.view(id!(empty_conversation)).set_visible(false);
-            }
-            State::ModelSelectedWithEmptyChat => {
-                self.view(id!(no_downloaded_model)).set_visible(false);
-                self.view(id!(no_model)).set_visible(false);
-                //self.view(id!(main)).set_visible(true);
-                self.view(id!(empty_conversation)).set_visible(true);
-            }
-            State::ModelSelectedWithChat => {
-                self.view(id!(no_downloaded_model)).set_visible(false);
-                self.view(id!(no_model)).set_visible(false);
-                self.view(id!(empty_conversation)).set_visible(false);
-            }
-        }
 
         // TODO: Rename "chat_history", "chat_count", etc, they are messages of a chat
         // can be confused with the actual Chat type.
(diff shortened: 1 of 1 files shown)


src/chat/chat_panel.rs
```diff
@@ -929,6 +869,86 @@ impl ChatPanel {
...
+
+        macro_rules! hide {
+            ($view_id:ident) => {
+                self.view(id!($view_id)).set_visible(false)
+            };
+        }
+
+        let store = scope.data.get_mut::<Store>().unwrap();
+
+        enum ComputedState {
```
noxware, line 888: Technically, visibility can be modified directly inside the `if` conditions but declaring a list of all possible states makes the code a bit more declarative and easier to understand.

src/chat/chat_panel.rs
```diff
@@ -929,6 +869,86 @@ impl ChatPanel {
             auto_scroll_cancellable: false,
         };
     }
+
+    fn update_visibilities(&mut self, scope: &mut Scope) {
+        macro_rules! show {
```
noxware, line 874: This tiny detail makes the imperative visibility code shorter inside this function :) A simple closure would be a bit longer because it would need to receive an `id!(...)` as an argument, also, it may cause issues because it would need to capture mutable references to `self`.

src/chat/chat_panel.rs
```diff
@@ -929,6 +869,86 @@ impl ChatPanel {
             auto_scroll_cancellable: false,
         };
     }
+
+    fn update_visibilities(&mut self, scope: &mut Scope) {
```
noxware, line 873: Keeping visibility controls in a single place makes the code much more readable and made me easy to debug all the issues :)

Although, this is just a small improvement. Ideally I believe view logic and event handling logic should be decoupled even further.

fmzbl: even though the function itself makes sense, i believe splitting the state is a mistake here, for a couple of reasons: first is that having two data structures without a real difference in what they represent is worse that centralizing them in one, they are both the state of the widget, why separating them? second is that you've flattened the state here, its not clear when the two enums exists at each point of the state of the widget, adding more cognitive load and problems in the future. by keeping it in the same one we could draw a graph with the state and its transitions more easily, making it easier for the widget to act over it now and better for maintainability.

src/chat/chat_panel.rs
fmzbl: not a big fan of this. im not against macros, but in this case it doesnt make much sense imo. its called `hide`, but it only hides views inside self, its not clear what it recieves, since the id becomes just a token that only makes sense when looking at the macro using it in onether macro (`id!`). leaving it exist only inside of this only function make it less bad, but still, seems like complexity without any real value inside of it, that is only use in this part of the project.

src/chat/chat_panel.rs
noxware: > two data structures without a real difference
As I mentioned in a comment, `ComputedState` is not necessary. It's just for organization. I can remove it if you prefer. But it's not the same as what is called "state" in the widget.

>  is worse that centralizing them in one,

Yes, this centralizes state. The state is centralized at the store. And is computed from there, instead of duplicating it (which is what `ChatPanelState` was doing). The `enum` is just for code organization and can be removed by putting the code directly in the corresponding if blocks (but that's less readable in my opinion). Maybe the world `state` is confusing you? 

`ChatPanelState` stores stuff that already exists in the store and can be computed from there, duplicating data unnecessary and provoking two sources of truths. 

Maybe with a little more effort I can get rid of `ChatPanelState` completely (instead of partially as I did) but that's out of the scope of this task.

noxware: https://github.com/moxin-org/moxin/commit/425f579a638512ee419e9db812478d3eaf3189ce this is how it would look without the helper enum. Probably looking it like this helps to resolve the confusion. This is 100% equivalent. There is absolutely no difference (just that it has some stuff duplicated, a lot of long nesting and is arguably less readable).

fmzbl: the point of ChatPanelState is to have a representation of the state of the widget, its not "data" as in the store. we can make the widget act over it, and make the transitions based on the store or whatever we want separately, without it we would be making boolean checks all over the place, like the ones you're doing in this function, or keeping single boolean variables. right now you're mixing both approaches, which is even worse.

src/chat/chat_panel.rs
noxware: I have no problem on changing this if someone else agrees 👍 I just wanted the code to look more clean and shorter by not needing to write `self.view(id!(SOMETHING)).set_visible(SOMETHING)` all over the place. But I know it functionally makes no difference.  I would like to know what someone else thinks before removing this 🙂

Anyways, clarifying specific parts of what you said:

> but it only hides views inside self

This is why it's defined just here and not as some kind of global utility.

>  its not clear what it recieves, since the id becomes just a token that only makes sense when looking at the macro

I don't know if I'm understanding this part correctly so please excuse me if my answer to this is not precise. But as far as I understand, this would happen with a closure as well but that is not a problem because it exists inside this function and you have the implementation visible.

fmzbl: you cannot just pass a bare token identifier to a clousure, you'd need to pass a &[LiveId], which is what id! gives u, though this is a less important detail.

fmzbl: i would just bind the views at the beginning and then do set_visible()

```rust
let empty_conversation_view = self.view(id!(empty_conversation)); 
...
match state {
    ...
    empty_conversation_view.set_visible(false)
```

and so on.

noxware: > you cannot just pass a bare token identifier to a clousure

Yes, that's why it's a macro and not a function. Thanks to that you can do `show!(something)` and not `show(id!(something)`.

> i would just bind the views at the beginning and then do set_visible()

Sounds good enough to me. Will change it to something like that then 👍

src/chat/chat_panel.rs
noxware: > the point of ChatPanelState is to have a representation of the state of the widget

But this is currently "app state", no "local state". Like is not state of the widget itself, but state from somewhere that the widget uses to compute it's view. 

100% proof of that this is no local state is that you can compute it completely. You could have a `state()` function that computes the widget state from the `store` returning some kind of `State` and that would work without storing it in the widget struct.

The widget struct  doesn't need to hold anything. Holding it may act as a kind of "cache" if computations are expensive, but that would be premature optimization.

> we can make the widget act over it, and make the transitions based on the store or whatever we want separately, without it we would be making boolean checks all over the place, like the ones you're doing in this function

Why do you want to act on two things instead of one? Like if you need to send something like `load_model` to the store, and at the same time transition the local state to `Idle`, it may happen that `load_model` fails and the store would not transition to the next state (thinking it like finite state machines) but you updated your local state and now you have 2 source of truth saying two different things.

And this is not an hypotetical case, look at what is in main for an example:

```
    fn load_model(&mut self, store: &mut Store, downloaded_file: DownloadedFile) {
        self.state = ChatPanelState::Idle;
        self.view(id!(main)).set_visible(true);
        self.view(id!(empty_conversation)).set_visible(true);
        self.view(id!(no_model)).set_visible(false);
        self.view(id!(no_downloaded_model)).set_visible(false);

        store.load_model(&downloaded_file.file);
    }
```

Here you transition the local state to `Idle`, but the `load_model` that comes after it can fail. If it fails, the store doesn't transition but your local state did, generating an inconsitency that would need help of god to debug. 

In general, keeping two structures in sync is highly error-prone no matter the kind of app or framework you are using and it can be a maintainability madness. If one "state" can be computed from other "state", it should be.

Of course, if the widget actually had some kind of owned state, then it would be ok to have it in the widget, but this widget doesn't.

So, I must insist, keeping track of this in the widget is a bad idea, and can be the source of worst state issues eventually.

src/chat/chat_panel.rs
joulei: I know this is resolved but adding to the conversation

+1 to the idea of avoiding unnecessary local macros. I like the idea of having a shortcut for visibility but I would explore it within Makepad itself (I'd like `hide` and `show` methods) or as a separate helper crate, so that we make it consistent throughout the app and our other apps as well.

I'd also add a +1 to the notion of making clear what `something` is in `show!(something)`. Also almost everything in Makepad that deals with `LiveId` looks like `animator_play(id!(hover.on));` or `self.view(id!(some_view))`, I think I'd be best to be consistent with that (or rethink it within Makepad).

src/chat/chat_panel.rs
```diff
@@ -712,40 +711,13 @@ impl WidgetMatchEvent for ChatPanel {
...
-
-        let jump_to_bottom = self.view(id!(jump_to_bottom));
-        match self.state {
-            ChatPanelState::Streaming {
-                auto_scroll_pending: true,
-                ..
-            } => {
-                // We avoid to show this button when the list is auto-scrolling upon
-                // receiving a new message. Otherwise, the button flicks.
-                jump_to_bottom.set_visible(false);
```
jmbejar, line 731: @noxware How this specific logic was translated to the new approach? I can understand everything else but this one seems lost.

Note: This logic seems to be not longer working on `main`, probably because the logic to detect if the user has scrolled up the messages list is fragile. But I wouldn't lost this logic although we would have to fix it later.

src/chat/chat_panel.rs
```diff
@@ -468,122 +471,43 @@ pub struct ChatPanel {
...
-    state: ChatPanelState,
+    state: State,
 }
 
 impl Widget for ChatPanel {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
         self.view.handle_event(cx, event, scope);
         self.widget_match_event(cx, event, scope);
+        self.update_state(scope);
+        dbg!(self.state);
```
noxware, line 482: I will leave this line here until someone else validates the new changes as it gives visibility of the state.

src/chat/chat_panel.rs
```diff
@@ -442,23 +445,23 @@ live_design! {
...
 }
 
-impl Default for ChatPanelState {
-    fn default() -> ChatPanelState {
-        ChatPanelState::Unload {
-            downloaded_model_empty: true,
-        }
+impl Default for State {
+    fn default() -> Self {
+        Self::Unknown
```
fmzbl, line 464: we can just derive Default and do #[default] over Unkown since the default is now a unit variant

src/chat/chat_panel.rs
```diff
@@ -594,105 +518,67 @@ impl Widget for ChatPanel {
...
+                self.redraw(cx);
+            } else if let ChatPanelAction::UnloadIfActive(file_id) = action.cast() {
+                if store
+                    .chats
+                    .get_current_chat()
+                    .map_or(false, |chat| chat.borrow().file_id == file_id)
+                {
+                    self.unload_model(cx);
+                    store.chats.eject_model().expect("Failed to eject model");
                 }
```
fmzbl, line 555: can we use a match here? we would avoid the else if

noxware: A single match is not possible because the types are different. This is also why we need to call `.cast()` multiple times.

The previous code did something like

```rust
match action.cast() {
  TypeA::SomeVariant => { ... }
  _ => {}
}

match action.cast() {
  TypeB::SomeVariant => { ... }
  _ => {}
}

match action.cast() {
  TypeC::SomeVariant => { ... }
  _ => {}
}
```

With a lot of `_ => {}` and single branch matches (I think there was one match with two branches, but only one of them)

So... those matches are equivalent to

```rust 
if let TypeA::SomeVariant = action.cast() {
  ...
}

if let TypeB::SomeVariant = action.cast() {
  ...
}

if let TypeC::SomeVariant = action.cast() {
  ...
}
```

Which is the more readable approach for me. And also, [this what clippy would recommend you to do](https://rust-lang.github.io/rust-clippy/master/index.html#/single_match)

I compacted the ifs with `else if` to avoid checking all conditions if one previously matched (as it can only match one of them), but I must agree that `else if` makes the thing look "tight".

Do you agree with not using `else if` but still use `if let`?

fmzbl: ooh ok, didn't realize that, its fine then yea

jmbejar (approved):
We'll be merging a bit later, but it is ready 👍 👏

PR #100, opened 2024-06-13
Save and load chats from disk

fmzbl:


src/data/store.rs
```diff
@@ -79,13 +80,37 @@ impl Store {
...
+    // TODO: This wrappers feel weird.
+    pub fn create_new_chat(&mut self, file: &File, set_as_current: bool) {
+        if self.chats.create_new_chat(file, set_as_current).is_ok() {
+            // TODO: This breaks, FIX THE PREFERENCES!!!
+            // self.preferences.set_current_chat_model(file.id.clone());
+        }
+    }
+
+    /// This loads the model
+    // TODO: Feels weird to do this wrappers
```
jmbejar, line 98: I would agree that simple wrappers do not make sense. But here we have the chance to have more than a wrapper.

As I said in another comment, this function should be responsible for getting the `available_model_files` (no real need to receive it from the caller) and call the `chats.set_current_chat` functions.

src/data/chats/mod.rs
```diff
@@ -75,22 +118,60 @@ impl Chats {
...
+        let file_id = self.get_current_chat().unwrap().borrow().file_id.clone();
+        let file = available_model_files
+            .iter()
+            .find(|file| file.id == file_id)
+            .expect("Attempted to start chat with a no longer existing file");
+
+        file
+    }
+
+    pub fn set_current_chat(
```
jmbejar, line 131: Ah ok.. Just to wrap up my previous comments here is a proposal to, I think, simplify things:

This function should be:

```rust
    pub fn set_current_chat(&mut self, chat_id: ChatID, file: &File) -> Result<()> {
        self.current_chat_id = Some(chat_id);

        if self
            .loaded_model_id
            .as_ref()
            .map_or(true, |m| *m != file.id)
        {
            self.load_model(file)?;
        }

        Ok(())
    }
```

And in `Store` we should have:

```rust
    pub fn set_current_chat(&mut self, chat_id: ChatID) {
        let available_files: Vec<File> = self
            .downloads
            .downloaded_files
            .iter()
            .map(|d| d.file.clone())
            .collect();

        let file_id = self.chats.get_current_chat().unwrap().borrow().file_id.clone();
        let file = available_files
            .iter()
            .find(|file| file.id == file_id)
            .expect("Attempted to start chat with a no longer existing file");

        if self
            .chats
            .set_current_chat(chat_id, &file)
            .is_ok()
        {
            // TODO: This breaks, FIX THE PREFERENCES!!!
             ....
            // self.preferences
            //     .set_current_chat_model(current_chat_model_file_id);
        }
    }
```

The main goal is to avoid passing around the `available_model_files` because there is no need for it.

Here is a complete diff for this idea: https://gist.github.com/jmbejar/0a94ae5c823f375bff081f45e9af65c2

@fmzbl what do you think?

src/data/chats/chat.rs
```diff
@@ -35,24 +41,72 @@ enum TitleState {
...
+
+        let (tx, rx) = channel();
+
+        Ok(Chat {
+            id,
+            model_filename,
+            file_id,
+            messages,
+            // TODO: This is kinda bad, we set the sender and reciever which need to
+            // be set correctly later
```
jmbejar, line 97: Looking at this comment I feel we shouldn't have our `Chat` struct implementing the serialization/deserialization to JSON. An alternative is to have functions converting `Chat` to/from `ChatData` (which can implement `Serde` easily), and make use of those functions in `load` and `save` functions. This idea has the benefit of removing all the `serde(skip)` added in the main struct.

In any case, I wouldn't block this PR for this. Just sharing how I would address this comment.

src/data/chats/chat.rs
```diff
@@ -71,9 +125,36 @@ impl Chat {
...
+    }
+
+    pub fn save(&self) {
+        let json = serde_json::to_string(&self).unwrap();
+        let path = self.chats_dir.join(self.file_name());
+        write_to_file(path, &json).unwrap();
+    }
+
+    // TODO: Make a field
+    fn file_name(&self) -> String {
```
jmbejar, line 154: Nit: Why would make a field if this is just a computed value? Maybe we can just put this calculation inline in the `save` function?

src/data/chats/chat.rs
```diff
@@ -207,9 +288,11 @@ impl Chat {
                 ChatTokenArrivalAction::AppendDelta(response) => {
                     let last = self.messages.last_mut().unwrap();
                     last.content.push_str(&response);
+                    self.save();
                 }
                 ChatTokenArrivalAction::StreamingDone => {
                     self.is_streaming = false;
+                    self.save();
                 }
             }
```
jmbejar, line 297: Nit: `self.save();` right after the match block.

src/data/store.rs
```diff
@@ -79,13 +80,37 @@ impl Store {
...
 
-    pub fn load_model(&mut self, file: &File) {
-        if self.chats.load_model(file).is_ok() {
-            self.preferences.set_current_chat_model(file.id.clone());
+    // TODO: This wrappers feel weird.
+    pub fn create_new_chat(&mut self, file: &File, set_as_current: bool) {
+        if self.chats.create_new_chat(file, set_as_current).is_ok() {
+            // TODO: This breaks, FIX THE PREFERENCES!!!
+            // self.preferences.set_current_chat_model(file.id.clone());
+        }
```
jmbejar, line 94: I would leave the existing `load_model` function like this:

```rust
    pub fn load_model(&mut self, file: &File, set_as_current: bool) {
        // We are creating a new chat because there is no other way to start a new conversation
        // with a model. This is a temporary solution until we have a better way to start a chat

        if self.chats.load_model(file).is_ok()
            && self.chats.create_new_chat(file, set_as_current).is_ok()
        {
            self.preferences.set_current_chat_model(file.id.clone());
        }
    }
```

As stated by the comment, the `create_new_chat` will be shortly removed with the PR I'm working on.

By the way, the preference line is working in my local setup.

src/data/chats/chat.rs
fmzbl: oh i never did this, its probably not that big of a deal, but this creates a new String on each save rn

src/data/chats/mod.rs
fmzbl: sure, this makes sense, I probably used Chats as a bigger "implementator" and didnt let store handle things, this way we can use the store more as the composer, so nice, it is probably better

fmzbl: yea looking at the code again makes sense to do it in the store, tough its a bit confusing that chats is public and sometime we interface with it and sometimes we dont from the outside, so having two public (tough one accessed through the chats field) set_current_chat that do diffrent things can be a bit confussing, but for now i feel like its better ye

jmbejar: The name duplication is a fair pain point. Renaming `Store.set_current_chat` to `Store.select_chat`, which is a name that makes sense from the caller point of view.

src/data/chats/chat.rs
jmbejar: I see..

fmzbl:
Continuing in #103.

PR #101, opened 2024-06-14
Support cargo-packager for packaging on macOS; add `moxin-runner`

kevinaboos:
Note that this also updates the `makepad-widgets` dependency to the latest version of makepad such that we can use the brand new `MAKEDPAD_PACKAGE_DIR` env var, which is required for packaging up resources into an app bundle and telling makepad's resource loader where to find them.

see commits and README.md for more details


jmbejar (dismissed):

juntao (approved):
LGTM

PR #112, opened 2024-06-20
feat: Change download location

ZhangHanDong:
Currently only tested on macOS, not yet tested on Linux.


kevinaboos:
In my view, the simple solution of adding an `on_completed()` callback fn to the download metadata struct is the right option. That way, the downloaded file can move itself to the newly-chosen Model Downloads directory after the download has completed.

ZhangHanDong:
@jmbejar @kevinaboos 

My solution is: After manually updating the download directory, send a notification to the backend main loop to automatically update the download directory. This way, it is not necessary to restart the application.

However, there is an issue: if a model is currently being downloaded, it will continue to download to the previous directory after the update. Only new downloads will use the new directory. At least this won’t affect the user experience.

jmbejar:
Great progress so far! 

However, I was testing and I found we have issues in the case you're downloading a file at the time of changing the models' folder. The issue actually happens if you pause/resume a download initiated before changing folder. The download will resume in the new location, but for some reason the application still points to the older folder when the model is completed and you want to use it (resulting in an error).

I'm digging into this issue right now, so I will let you what would be a good fix for this.

jmbejar:
> I'm digging into this issue right now, so I will let you what would be a good fix for this.

@ZhangHanDong Based in your last message in Discord, I assume you're around it. So ignore my last sentence :) Let me know if you need help.

jmbejar (approved):
Tested again 👏  Thanks for adding the fix to the last reported issue!

PR #117, opened 2024-06-24
fix(backend): update wasm for support utf-8

L-jasmine:
In some models, when generating emojis, the full UTF-8 string is not generated all at once, so it is necessary to wait for multiple tokens to be returned and then combine them into a complete UTF-8 string before returning it.

Here are the specific changes of the wasm:

https://github.com/L-jasmine/LlamaEdge/commit/e83d23916e59c5d613f4611c935f199a113a846a


jmbejar (approved):
👏 👏 👏

PR #122, opened 2024-06-26
Use Makepad Button as much as possible

jmbejar:
This PR completes the task of making use of the official Button widget, instead of our previous approach which was to rely on RoundedView and basic click events


src/chat/chat_panel.rs
```diff
@@ -33,6 +33,16 @@ live_design! {
     ICON_STOP = dep("crate://self/resources/icons/stop.svg")
     ICON_JUMP_TO_BOTTOM = dep("crate://self/resources/icons/jump_to_bottom.svg")
 
+    RoundedButton = <MoxinButton> {
```
noxware, line 36: nitpick: Technically, MoxinButton is also rounded, right? Idk if there is a better name for this.

src/chat/chat_panel.rs
```diff
@@ -147,39 +157,45 @@ live_design! {
...
                 }
-                icon_walk: {width: 12, height: 12}
+            }
+            icon_walk: {
+                margin: {top: 6, left: -4},
             }
         }
     }
 
+    PromptRoundedButton = <RoundedButton> {
```
noxware, line 183: nitpick: Related to my previous comment, if the RoundedButton name changes this should change, but in this case I think the name could be as easy as `PromptButton`.

src/chat/chat_panel.rs
```diff
@@ -147,39 +157,45 @@ live_design! {
...
+        draw_bg: {
+            radius: 6.5,
+            color: #D0D5DD
+        }
+        icon_walk: {
+            margin: {top: 0, left: -4},
+        }
+
+        padding: {right: 4},
+        margin: {bottom: 2},
```
noxware, line 196: I see `RoundedButton` already has:

```
padding: {right: 4},
margin: {bottom: 2},
```

So I think it's safe to omit them.

noxware (approved):
LGTM

Left very minor comments that are not important

PR #124, opened 2024-06-26
Use `robius-url-handler` to handle incoming URLs with scheme `moxin://`

kevinaboos:
Tested working on Debian-like Linux and macOS.

Note that this functionality will only work for properly packaged application bundles (i.e., using cargo-packager) that have been installed on your system, e.g., using `sudo dpkg -i` on Linux or installing the app bundle via the `moxin-*.dmg` disk image on macOS.

On Linux, the handled URLs are appended to the file `/tmp/moxin_incoming_url.txt`.
On macOS, that file is located at `$TMPDIR/moxin_incoming_url.txt`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,6 +170,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block2"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jmbejar (approved):
Thanks for adding support to URL handlers 👏 

We will define soon how we should use this feature in Moxin, probably triggering a model download from HuggingFace.

PR #129, opened 2024-06-28
Auto-determine dependencies for Linux packaging

kevinaboos:
These are now specified for Debian and Pacman packages.

Uses a new version of cargo-packager, as now specified in the README.

Also fixes AppImage packages to _always_ work even on a fresh Linux install.


jmbejar (approved):
🤩

PR #133, opened 2024-07-01
Chat Inference Params

noxware:
## Changes

- Adds a toggle panel to configure the inference params of the chat.
- Adds customized slider and switch widgets.
- Make the existing chat panel `Fill` so it adapts better to the new size constraints.

## Recording

https://github.com/moxin-org/moxin/assets/7684329/35158887-a0af-49b9-b5c9-30c8b501f088

## Notes for reviewers

- Figma design [here](https://www.figma.com/design/94gxq1QD6WTaNDJh1SGWsh/MoxinUI---New?node-id=44-2272&t=aPeUkwpw7P2wYzOv-4).
- Most relevant file is `chat_params.rs` which is the panel itself.


src/chat/chat_params.rs
```diff
@@ -0,0 +1,289 @@
...
+    import crate::shared::widgets::MoxinButton;
+    import crate::shared::widgets::MoxinSlider;
+    import crate::shared::widgets::MoxinSwitch;
+    import crate::shared::widgets::MoxinTextInput;
+    import makepad_draw::shader::std::*;
+
+
+    // TODO: Use proper icons for right panel.
+    ICON_CLOSE_PANEL = dep("crate://self/resources/icons/open_left_panel.svg")
+    ICON_OPEN_PANEL = dep("crate://self/resources/icons/close_left_panel.svg")
```
noxware, line 20: I tried copying as svg [this](https://www.figma.com/design/LIaqW5HF3CGtq4XMpDxDMD/MoxinUI---LLM-Desktop-App?node-id=1885-2756&t=WFGkMXe5bBlBHE3I-4) but it blows as shown in the following video, so I'm using the same icons that are used on the left side.

https://github.com/moxin-org/moxin/assets/7684329/ce14d91d-c37c-4406-8d53-e9d94cca45f5

src/chat/chat_params.rs
```diff
@@ -0,0 +1,289 @@
...
+                        max: 2.0
+                    }
+
+                    top_p = <MoxinSlider> {
+                        text: "Top P"
+                        min: 0.0
+                        max: 1.0
+                    }
+
+                    <View> {
```
noxware, line 102: Idk if there is a better way to have the text on the left and the switch on the right without a view.

src/chat/chat_params.rs
```diff
@@ -0,0 +1,289 @@
...
+                                text_style: <BOLD_FONT>{font_size: 10},
+                                color: #000
+                            }
+                            text: "Stop"
+                        }
+                        stop = <MoxinTextInput> {
+                            width: Fill,
+                            // TODO: This should be something like min-height, allowing
+                            // the text input to grow.
+                            height: 65,
```
noxware, line 143: - Without a way to use min-height, the text will overflow after entering some lines of text (like 4 lines).
- We can set this to `Fit` to avoid the issue but that would not match the design.
- We can "mitigate" a bit this by asking for a comma-separated list of terms instead of new line separated list.

jmbejar: A comma-separated list of terms sounds good to me as a first implementation. I feel this will be improved having a more complex UI to add each team separately (like adding labels)

src/chat/chat_params.rs
```diff
@@ -0,0 +1,291 @@
...
+        top_p.set_value(ip.top_p.into());
+        max_tokens.set_value(ip.max_tokens.into());
+        frequency_penalty.set_value(ip.frequency_penalty.into());
+        presence_penalty.set_value(ip.presence_penalty.into());
+        stop.set_text(&ip.stop);
+
+        // Idk why, but without the check, this binding will not work when the screen is a non empty chat.
+        // I don't get the relation, maybe is something related to the mandatory `ctx` parameter or the
+        // animator.
+        if stream.selected(cx) != ip.stream {
```
jmbejar, line 235: I presume this is necessary to avoid a situation where you are firing the toggle animation in this `draw_walk`, which cases to trigger redraws, so the animation would be restarted again (without having the change to make any progress).

Maybe it is better to implement this "data binding" relying more in retained mode? I just tested the following and seems to work:

```rs
        if let Event::Startup = event {
            let store = scope.data.get::<Store>().unwrap();
            let ip = &store.chats.inferences_params;

            let temperature = self.slider(id!(temperature));
            let top_p = self.slider(id!(top_p));
            let max_tokens = self.slider(id!(max_tokens));
            let frequency_penalty = self.slider(id!(frequency_penalty));
            let presence_penalty = self.slider(id!(presence_penalty));
            let stop = self.text_input(id!(stop));
            let stream = self.check_box(id!(stream));

            temperature.set_value(ip.temperature.into());
            top_p.set_value(ip.top_p.into());
            max_tokens.set_value(ip.max_tokens.into());
            frequency_penalty.set_value(ip.frequency_penalty.into());
            presence_penalty.set_value(ip.presence_penalty.into());
            stop.set_text(&ip.stop);

            stream.set_selected(cx, ip.stream);
        }

        ...
    }
    
        self.view.draw_walk(cx, scope, walk)
    }
```

It is not a situation where the settings values could be changed from other sources, so it feels safe to make the local state source of truth after taking the data from the store at startup time.

Even thinking in the case of having this setting per-conversation, the form data can be easily repopulated from the store just reacting to the existing `ChatHistoryCardAction::ChatSelected` action.

jmbejar (commented):
Is it possible that the slider text inputs are not working? I can sort of delete the numbers but I can not enter numbers manually at all.

src/chat/chat_params.rs
```diff
@@ -0,0 +1,291 @@
...
+                        padding: {left: 4}
+                        <Label> {
+                            draw_text: {
+                                text_style: <BOLD_FONT>{font_size: 10},
+                                color: #000
+                            }
+                            text: "Stream"
+                        }
+                        <Filler> {}
+                        stream = <MoxinSwitch> {}
```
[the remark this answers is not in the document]
jmbejar: Found the missing part, here is a fix: https://github.com/moxin-org/moxin/pull/135

src/shared/widgets.rs
```diff
@@ -346,4 +346,104 @@ live_design! {
...
+                sdf.fill(pill_color_on);
+
+                // Right side of the pill
+                sdf.circle(self.rect_size.x - pill_radius, pill_radius, pill_radius);
+                sdf.fill(pill_color_off)
+                sdf.blend(self.selected);
+                sdf.fill(pill_color_on);
+
+                // The union/middle of the pill
+                sdf.rect(pill_radius, 0.0, self.rect_size.x - 2.0 * pill_radius, self.rect_size.y);
```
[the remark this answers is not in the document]
jmbejar: Playing a bit I discovered it seems to be a rect border. Setting the stroke to 0 didn't well as I expected for some reason, so I changed to avoid using `blend`.

The following works for me:

```
                // Left side of the pill
                sdf.circle(pill_radius, pill_radius, pill_radius);
                sdf.fill(mix(pill_color_off, pill_color_on, self.selected))

                // Right side of the pill
                sdf.circle(self.rect_size.x - pill_radius, pill_radius, pill_radius);
                sdf.fill(mix(pill_color_off, pill_color_on, self.selected))

                // The union/middle of the pill
                sdf.rect(pill_radius, 0.0, self.rect_size.x - 2.0 * pill_radius, self.rect_size.y);
                sdf.fill(mix(pill_color_off, pill_color_on, self.selected))
```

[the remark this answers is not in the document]
noxware: Thank you so much! I was using blend because I started from [this reference](https://github.com/project-robius/makepad_taobao/blob/642990a4f0aefa5a319ee3e55827e44b763ae539/src/settings/settings_screen.rs#L35). I didn't suspect it was causing the issue. I expected blend to do some overlapping of thing but it may be doing something I didn't expect.

src/chat/chat_params.rs
noxware: > You are firing the toggle animation in this draw_walk, which cases to trigger redraws, so the animation would be restarted again

Yeah, I had that theory for the checkbox widget when I experienced this [other thing](https://github.com/makepad/makepad/pull/485#discussion_r1663268371) in the slider one. Although, I still don't understand why it worked on empty chats. 

I probably had to update that comment at least saying "this is because setting the value also activates the animation as a side effect causing visual issues so we avoid calling it".

> Maybe it is better to implement this "data binding" relying more in retained mode? I just tested the following and seems to work:

I tried the same when I hit the switch problem the first time. But I discarded the commit because of the following:

- We are changing the approach just because one and only one widget is miss behaving.
  - Fixed with a simple if (2 harmless lines).
- We lose proper data binding in exchange of incomplete data binding.
  - Although, as you mention, it works in this current case, now everyone must know to NOT mutate this from the store directly or they would introduce a bug with not immediate visual implications (risky). Yeah we can put a comment saying "hey, this comes from X widget, do not mutate this directly" but we can also not have that risky point of failure on the first place.
- That solution, or the `if` solution. Both are workarounds. Both are just required because of the switch. Both are not required otherwise. And one involves less functional changes and is more robust than the other.
- If we want retained mode, we don't need data binding on the first place, we should directly ask the widget it's data when we need it. Something like one of the following approaches:

image

- And I tried doing the previous point, to have no-dirty, no-fragile, robust retained mode UI, but I stopped in the middle because I was like... "hey... I'm seriously changing my whole implementation just because of this dumb switch...?"
  - ...Thanks god I found how to fix the switch.

So, this would be my course of action:
- Let's leave the thing how it's currently done. I don't see any problem with the implementation. `Don't touch it. It works.` said a sage... somewhere... probably...
  - But let's update this comment giving the animation explanation.
- Unless
  - We prefer going fully retained as an agreement.
  - We found an undetected, not easy to resolve issue with this and we MUST go retained, in which case:
    - I would prefer the approach from the diagrams.
    - But... I'm lazy... So I would actually pick your suggestion and add a warning comment in the store and call it a day.

noxware:
> Is it possible that the slider text inputs are not working?

@jmbejar I tried at `c0980c86d` of the makepad repo (before rik merged by branch) with a simple example but it doesn't work either.

https://github.com/moxin-org/moxin/assets/7684329/d1d7e352-cd7a-4997-8163-72a1c4314057

So rest assured, we didn't introduce any issue :P

src/chat/chat_params.rs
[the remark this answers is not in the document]
noxware: https://github.com/moxin-org/moxin/assets/7684329/b112066e-fea6-4ca8-849a-50f9001f67f5

Tested and working

src/chat/chat_params.rs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        self.view.draw_walk(cx, scope, walk)
+    }
+}
+
+impl WidgetMatchEvent for ChatParams {
+    fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
+        let store = scope.data.get_mut::<Store>().unwrap();
+        let ip = &mut store.chats.inferences_params;
```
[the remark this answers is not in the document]
noxware: Oh sorry, I didn't understood well last time then.

I implemented per chat settings, the change is quick and small. https://github.com/moxin-org/moxin/pull/133/commits/ca1567bcb45a8b3c2dec86a8f466312933d59b7a

Screenshot 2024-07-04 at 19 13 57

https://github.com/moxin-org/moxin/assets/7684329/919417e3-c68b-4bb9-b671-555cf6a8106e


But I had to track when the chat changes to ask makepad for a redraw (check the commit), and I don't like that. Do you have a different idea?

jmbejar (approved):

PR #134, opened 2024-07-03
Switch loaded model upon chat change

jmbejar:
When the user switches to another chat conversation, the last used model is automatically loaded.

In addition, there are two bug fixes:

* Save & Regenerate buttons were visible for AI messages, which was wrong (issue introduced when changed to use Makepad buttons).
* Fixed weird issue when chat messages were not appearing at all (the problems was the scroll position of the messages PortalList when coming from a longest conversation).

- [x] Check if changes would break existing users because of the chat persistence.
Note about existing users: Chat conversations persisted with earlier versions of the app won't auto-select the last used model because this piece of data was never stored. We could use the model used in the first interaction but it doesn't seem to worth adding more code to cover such case. As far the conversation are updated, they will be persisted in our new format and the app will behave in the new way.

### DIFF
--- a/src/data/store.rs
+++ b/src/data/store.rs
@@ -8,7 +8,6 @@ use chrono::{DateTime, Utc};
 use makepad_widgets::{DefaultNone, SignalToUI};
 use moxin_backend::Backend;
 use moxin_protocol::data::{Author, DownloadedFile, File, FileID, Model, ModelID, PendingDownload};
-use std::path::PathBuf;
 use std::rc::Rc;
 
 pub const DEFAULT_MAX_DOWNLOAD_THREADS: usize = 3;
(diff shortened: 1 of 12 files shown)


src/data/store.rs
```diff
@@ -94,64 +93,65 @@ impl Store {
...
-                self.chats.create_empty_chat_with_model_file(file);
+                self.chats.create_empty_chat();
             }
         }
     }
 
     pub fn select_chat(&mut self, chat_id: ChatID) {
         self.chats.set_current_chat(chat_id);
+
+        if let Some(file_id) = self.get_last_used_file_id_in_current_chat() {
```
noxware, line 104: nitpick: maybe we can extract this chunk of added code to a function called `self.load_last_model_for_current_chat()` and call it here. Just for the sake of having smaller and self documented functions.

noxware (commented):
lgtm

left two small unimportant comments

noxware (dismissed):

PR #137, opened 2024-07-05
Delete Chat Modal

noxware:
# Changes

- Adds a confirmation modal when trying to delete a chat.
  - The modal is almost a copy of the existing modal for deleting models.
- [Design](https://www.figma.com/design/LIaqW5HF3CGtq4XMpDxDMD/MoxinUI---LLM-Desktop-App?node-id=816-4068&t=enlxvJ4uaQS0AlOe-4)
  
# Recording

https://github.com/moxin-org/moxin/assets/7684329/deee9d34-07cc-4472-829e-496a51cf1c53

### DIFF
--- a/src/chat/delete_chat_modal.rs
+++ b/src/chat/delete_chat_modal.rs
@@ -1,0 +1,217 @@
+use makepad_widgets::*;
+
+use crate::{
+    chat::chat_panel::ChatPanelAction,
+    data::{chats::chat::ChatID, store::Store},
+    shared::portal::PortalAction,
+};
+
+live_design! {
+    import makepad_widgets::base::*;
+    import makepad_widgets::theme_desktop_dark::*;
+    import makepad_draw::shader::std::*;
+
+    import crate::shared::styles::*;
+    import crate::shared::widgets::MoxinButton;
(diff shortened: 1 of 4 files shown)


src/chat/delete_chat_modal.rs
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        if self
+            .button(id!(wrapper.body.actions.delete_button))
+            .clicked(actions)
+        {
+            let store = scope.data.get_mut::<Store>().unwrap();
+            store.delete_chat(self.chat_id);
+            cx.widget_action(widget_uid, &scope.path, PortalAction::Close);
+            cx.redraw_all();
```
noxware, line 199: Tell makepad to fully redraw to incorporate the changes from the store.

joulei: I don't love this, but I get that you're avoiding the creation of some kind of action to listen to and redraw in the relevant parts of the UI. 
I think is fine for now, eventually we could add metadata to the `PortalAction::Close` action (so we listen to the right one) and redraw based on that.

joulei (approved):

PR #139, opened 2024-07-08
Avoid redraw call in render logic in Chat

jmbejar:
I was debugging the Chat screen and confirmed we had a `self.redraw(cx)` in a function that was called (indirectly) from a `draw_widget` function body. This change should not affect anything, but let's confirm with more testing.


fmzbl (approved):
didn't encounter any issues

PR #141, opened 2024-07-08
Last selected chat memory

noxware:
# Changes
- Keep track of the last time a chat has been selected.
- Load the last selected chat on app start.
- Bonus: When deleting the chat, move to the previous one you had selected.
  - This change lets me remove the last created chat id function that I think was only being relevant for this arbitrary behavior.
  
# Recording

https://github.com/moxin-org/moxin/assets/7684329/efc61e45-d8f1-4b1b-a69a-6a598d459482

### DIFF
--- a/src/data/chats/chat.rs
+++ b/src/data/chats/chat.rs
@@ -7,7 +7,6 @@ use moxin_protocol::protocol::Command;
 use std::path::PathBuf;
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::thread;
-use std::time::{SystemTime, UNIX_EPOCH};
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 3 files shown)


src/data/chats/chat.rs
```diff
@@ -51,6 +51,8 @@ struct ChatData {
     title: String,
     #[serde(default)]
     title_state: TitleState,
+    #[serde(default)]
```
noxware, line 54: chrono will pick the unix epoch as the default, and that's important for this to work in a backward compatible way.

src/data/chats/chat.rs
```diff
@@ -343,4 +349,8 @@ impl Chat {
             .unwrap();
         self.messages.truncate(message_index);
     }
+
+    pub fn touch_last_selected_at(&mut self) {
+        self.last_selected_at = chrono::Utc::now();
+    }
```
fmzbl, line 355: I think this naming is flawed. We are on the Chat struct, the concept of "touch" doesn't really exist at this level of abstraction, it also couples it too much with the caller. Furthermore, i feel like saving "selection" is a bit weird, why would a chat need to remember if it was selected or not? that's a caller/UI level concept. wouldn't it be better to save a last updated or something like that, and interpret a selection as a form of update from the outside? If anything i would keep the idea of a selected chat in the Chats struct and above, does this make sense?

src/data/chats/chat.rs
fmzbl: We're using different ways of handling dates now, one for setting the id and another one for this, can we unify it?

fmzbl (approved):

PR #145, opened 2024-07-10
Fix jump to bottom flicker

noxware:
# Problem

When you try to scroll further than the maximum possible amount, the jump to bottom button flickers.

https://github.com/moxin-org/moxin/assets/7684329/01148994-fe77-4947-b8bf-e0b84f7edc9f

# Changes

Detect if the end of the list has been reached in a different way. By telling the portal list to render one more item, and when the portal list renders that, flag that the portal list reached the end.

### DIFF
--- a/src/chat/chat_panel.rs
+++ b/src/chat/chat_panel.rs
@@ -395,6 +395,7 @@ live_design! {
 
                     UserChatLine = <UserChatLine> {}
                     ModelChatLine = <ModelChatLine> {}
+                    EndOfChat = <View> {height: 0.1}
                 }
 
                 <JumpToButtom> {}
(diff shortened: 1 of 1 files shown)


src/chat/chat_panel.rs
```diff
@@ -429,6 +429,9 @@ pub struct ChatPanel {
 
     #[rust]
     state: State,
+
+    #[rust]
+    portal_list_end_reached: bool,
```
noxware, line 434: I will leave this separated as it's just to support the portal list workaround here.

src/chat/chat_panel.rs
```diff
@@ -395,6 +395,7 @@ live_design! {
 
                     UserChatLine = <UserChatLine> {}
                     ModelChatLine = <ModelChatLine> {}
+                    EndOfChat = <View> {height: 0.1}
```
noxware, line 398: We still need to render something real with non-zero height at the end to make this trick work.

jmbejar (approved):

PR #149, opened 2024-07-17
[Backend] bump wasmedge version to 0.14.0 for support windows

L-jasmine:
In Wasmedge 0.13.5, there is a bug in the plugins mechanism on Windows, so Moxin needs to upgrade to Wasmedge 0.14.0 to support Windows.

When build on Windows, you may encounter the following error. This is because `wasmedge-sys` use `bindgen`, which requires clang. To resolve this error, you need to download LLVM and specify the environment variable `LIBCLANG_PATH`.
```
--- stderr
  thread 'main' panicked at G:\.cargo\registry\src\index.crates.io-6f17d22bba15001f\bindgen-0.69.4\lib.rs:622:31:
  Unable to find libclang: "couldn't find any valid shared libraries matching: ['clang.dll', 'libclang.dll'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"
```
It is only needed when running the build.rs script, so it only exists during the compilation phase.

### DIFF
--- a/moxin-backend/Cargo.toml
+++ b/moxin-backend/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 [dependencies]
 moxin-protocol = { path = "../moxin-protocol" }
 chrono = "0.4"
-wasmedge-sdk = { version = "=0.13.5-newapi", default-features = false, features = [
+wasmedge-sdk = { version = "0.14.0", default-features = false, features = [
     "wasi_nn",
 ] }
 log = "0.4.21"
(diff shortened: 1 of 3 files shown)


L-jasmine:
But I found that previous versions could run on Windows, whereas the current version cannot pop up the UI on Windows.

jmbejar (approved):

MrBeanCpp:
After updating `wasmedge` (system) from `0.13.5` to `0.14.0`, we may need to install `WASI-NN` plugin manually.
[Install and uninstall WasmEdge | WasmEdge Developer Guides](https://wasmedge.org/docs/start/install/#wasi-nn-plug-in-with-ggml-backend)

PR #150, opened 2024-07-17
add chat options menu

fmzbl:


jmbejar (commented):
This happens when collapsing the left sidebad, when the menu is still visible. This issue sounds similar to the one we had when navigating to other parts of the app

image

jmbejar:
@fmzbl Found a solution to the problem dismissing the floating menu from certain regions of the screen. Can you check https://github.com/moxin-org/moxin/pull/150/commits/004cb397d3c738688b44097a842e0e4581d2983c?

jmbejar (dismissed):
There are a bit of cosmetic issues I'd address before merging, but in general it is working now as expected 👍

fmzbl:
> @fmzbl Found a solution to the problem dismissing the floating menu from certain regions of the screen. Can you check [004cb39](https://github.com/moxin-org/moxin/commit/004cb397d3c738688b44097a842e0e4581d2983c)?

Yeah cool, that should fix the panel and navigating issues, we can remove that stuff then.

jmbejar (approved):

PR #157, opened 2024-07-19
Extract moxin panel (further)

noxware:


src/shared/toggle_panel.rs
```diff
@@ -71,6 +88,16 @@ impl Widget for TogglePanel {
...
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        if !self.initialized {
+            self.initialized = true;
+            // Ensure the progress is consistent with the state of the animator.
+            self.animator_panel_progress = if self.is_open(cx) { 1.0 } else { 0.0 };
+        }
+
+        let size_range = self.open_size - self.close_size;
+        let size = self.close_size + size_range * self.animator_panel_progress;
+        self.apply_over(cx, live! {width: (size)});
```
noxware, line 99: Problem: This has no effect on the first draw, unless I move it to `handle_event` callback where it would run for any event. 


https://github.com/user-attachments/assets/bf0d043d-0654-40a9-ac1f-3b7126c40974

jmbejar: It feels all this logic should be moved to the event handler. Specially to ensure we apply over the value only when animating, and also on initialization. I bet it could solve also this issue.

```rust
        if !self.initialized {
            self.initialized = true;
            if self.is_open(cx) {
                self.animator_panel_progress = 1.0;
                self.apply_over(cx, live! {width: (self.open_size)});
            } else {
                self.animator_panel_progress = 0.0;
                self.apply_over(cx, live! {width: (self.close_size)});
            };
            self.redraw(cx);
        }

        if self.animator_handle_event(cx, event).must_redraw() {
            let size_range = self.open_size - self.close_size;
            let size = self.close_size + size_range * self.animator_panel_progress;
            self.apply_over(cx, live! {width: (size)});
            self.redraw(cx);
        }

    }
```

Also, there is a Event::Startup event that could be used to avoid having the `initialized` flag, it is something I would try to see if it improve things in the code.

src/shared/toggle_panel.rs
```diff
@@ -0,0 +1,288 @@
...
+            instance opacity: 1.0
+
+            fn pixel(self) -> vec4 {
+                let color = sample2d_rt(self.image, self.pos * self.scale + self.shift) + vec4(self.marked, 0.0, 0.0, 0.0);
+                return Pal::premul(vec4(color.xyz, color.w * self.opacity))
+            }
+        }
+    }
+
+    CustomButton = <Button> {
```
noxware, line 22: Just a copy of MoxinButton to match moxin styles when porting this outside, as discussed internally.

fmzbl (approved):

jmbejar (approved):

PR #166, opened 2024-07-23
Smooth model switch

jmbejar:


moxin-backend/src/backend_impls/chat_ui.rs
```diff
@@ -316,22 +316,15 @@ fn create_wasi(
     file: &DownloadedFile,
     load_model: &LoadModelOptions,
 ) -> wasmedge_sdk::WasmEdgeResult<WasiModule> {
-    let ctx_size = if load_model.n_ctx > 0 {
```
jmbejar, line 319: Context: This change was originally done here https://github.com/moxin-org/moxin/blob/1056e1a0e2789f99bf7fc5ca40149b462a79c997/moxin-backend/src/backend_impls/api_server.rs#L38-L46, which is the currently active backend.

Making the change here as well since I'm removing those values from the protocol, but it does not have real effect right now.

src/chat/model_selector_loading.rs
```diff
@@ -0,0 +1,133 @@
...
+    import makepad_widgets::theme_desktop_dark::*;
+    import makepad_draw::shader::std::*;
+
+    import crate::shared::styles::*;
+    import crate::shared::widgets::*;
+    import crate::landing::model_card::ModelCard;
+
+    ANIMATION_SPEED = 1.5;
+
+    Bar = <RoundedView> {
```
jmbejar, line 14: There is a bunch of repeated logic with other loading widget implemented, but decided to not overcomplicate this PR at this point.

PR #170, opened 2024-07-25
README: add instructions for building and running on Windows x64

kevinaboos:
With these instructions, Moxin builds and runs but does not yet properly work when loading a model to chat with.

There is still a problem with WasmEdge or Moxin on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ data.*
 .vscode
 /dist
 /wasmedge
+/WasmEdge*
 .DS_Store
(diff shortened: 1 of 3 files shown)


hydai:
Hi @kevinaboos 
Could you please confirm that these two pre-shipped wasmedge tarballs(see the following screenshot) with the wasmedge library(LLVM disabled) and that the ggml plugin still needs the LLVM dependencies? IIRC, we built these two assets with `-DWASMEDGE_USE_LLVM=OFF` flag, so it should not require the LLVM libraries.

CleanShot 2024-07-25 at 14 17 16

These four wasmedge core tarballs will require the LLVM dependencies. So I don't think you need to use them
CleanShot 2024-07-25 at 14 31 45

kevinaboos:
> Hi @kevinaboos Could you please confirm that these two pre-shipped wasmedge tarballs(see the following screenshot) with the wasmedge library(LLVM disabled) and that the ggml plugin still needs the LLVM dependencies? IIRC, we built these two assets with `-DWASMEDGE_USE_LLVM=OFF` flag, so it should not require the LLVM libraries.

LLVM is needed for libclang, which is required by wasmedge-sys in order to generate bindings (and potentially do other things, I'm not really sure). This is only a build-time dependency, not needed at runtime.

Is there a way to configure wasmedge-sys to not need clang?

jmbejar (approved):

PR #171, opened 2024-07-25
[Backend] fix windows

L-jasmine:
Unable to access `http://0.0.0.0:8080/echo` on Windows, but it works on Linux. This causes it to not works on Windows.


kevinaboos (approved):
Confirmed working on Windows. Thanks!

PR #172, opened 2024-07-25
[backend] remove proxy when request api-server

L-jasmine:
Sometimes we need to set the http_proxy, but incorrect settings can affect access to api-server.wasm. So I’ve configured it so that no proxy is required when accessing api-server.wasm.


kevinaboos (approved):
confirmed working on Windows. Thanks!

PR #177, opened 2024-07-25
Assistant message alignment

noxware:
Screenshot 2024-07-25 at 12 37 35

### DIFF
--- a/src/chat/chat_line.rs
+++ b/src/chat/chat_line.rs
@@ -110,7 +110,7 @@ live_design! {
     ChatLineBody = <View> {
         width: Fill,
         height: Fit,
-        spacing: 12,
+        spacing: 20,
         flow: Down,
 
         sender_name_layout = <View> {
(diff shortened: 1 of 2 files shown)


src/chat/chat_panel.rs
```diff
@@ -122,6 +122,7 @@ live_design! {
         main_section = {
             body_section = {
                 bubble = {
+                    padding: {left: 0, bottom: 0, top: 0}
```
noxware, line 125: Most of bubble spacing is removed for assistant messages. Although the right padding was assumed when placing the jump to bottom button, so that's preserved.

src/chat/chat_line.rs
```diff
@@ -110,7 +110,7 @@ live_design! {
     ChatLineBody = <View> {
         width: Fill,
         height: Fit,
-        spacing: 12,
+        spacing: 20,
```
noxware, line 113: This is an extra adjustment. Without the bubble, 12 as spacing between the assistant name and the message may look to tight, so I added a bit more.

jmbejar (approved):
Thanks!

Issue #178, opened 2024-07-25
Mouse cursor does not change back to regular pointer after exiting the button or link area


guofoo:
When the mouse arrow-pointer is hovering over buttons and links it changes into a hand-pointer.

After it leaves the clickable area, the pointer should change back to arrow. But this does not happen right away for most buttons and links. Most of them they change only after leaving the "containing panel" area.
Problem ares: Resources Links, Download buttons, Chat with Model, Resume Chat, Info, and Delete buttons, etc.

Exceptions are the Show All Files / Only Recommended Files buttons, here the pointer behaves properly. 
The Chat hide panel and new chat buttons are ok, same with Discover, Chat and My Models screen buttons.

kevinaboos:
Echoing the same problem for Robrix too. Hover-exit actions don't seem to be triggering at all.

jmbejar:
This is much of a problem in Makepad, as we discussed in recent meetings. I'm currently exploring some possible solutions but there is a chance that we should wait for fundamental changes from Makepad mouse/touch events handling system.

jmbejar:
Most of the issues should now be fixed in `dev` branch, so let's test a bit and then close the issue once it is merged to `main`.

There is one situation where the cursor does not change back to the arrow after hovering a control. It happens with the Sort options (dropdown) but I confirmed it is an issue in all of the Makepad applications and examples. We should address this specific issue in the Makepad dropdown widget.

PR #180, opened 2024-07-26
fixed Issue 131

ZhangHanDong:
The overall idea is to use Store to share the state of opening and closing the info window (show_info), and then redraw the info button background based on this state.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -226,6 +226,7 @@ impl MatchEvent for App {
                     self.store.search.sort_models(criteria);
                 }
                 StoreAction::SetShowInfo(file_id, show_info) => {
+                    eprintln!("file_id in App {:?}", file_id);
                     self.store.set_show_info(file_id, show_info);
                     cx.redraw_all();
                 }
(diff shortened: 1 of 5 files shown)


joulei:
Hi @ZhangHanDong, thank you for looking into this.

While these changes mostly work, the main problem is that the button state is not updated (and these changes don't fix that yet, so you see some conflicts with mouse events afterwards). 
This happens because when the Modal is open, the mouse events that would normally cause a "hover-out" of the icon, are not being propagated downwards (which is generally what we want, so that the cursor does not interact with elements behind the modal while it's open).

I'll take a look at how we can fix this within the modal or button widgets (or maybe extending the event propagation system in Makepad). Because I think app developers shouldn't need to deal with updating that state.

That said, if we were to update the button state on our own, I'd perhaps use the modal closing action `PortalAction::Close` (with some metadata), and listen to it from the invoker, and based on that update status of the button.
This way we would avoid saving those values in the store for each button (this happens also with other modals). I'd avoid having UI visibility specifics in the Store.
Another simpler option could be that when opening the modal we immediately tell the `ButtonRef` (that we just used to check for .clicked), to update its hover state.
```rs
    if self.button(id!(info_button)).clicked(actions) {
           ...
            cx.widget_action(
                widget_uid,
                &scope.path,
                PortalAction::ShowPortalView(live_id!(modal_model_info_portal_view)),
            );
            self.button(id!(info_button)).set_hover(false) 
            // ^ something like this, again not ideal to deal with as an app dev
        ...
    }
```
Either way for those options we would have to extend the button API in Makepad to allow overriding that state (I'm not sure if we want to do that).

Lastly, this other issue https://github.com/moxin-org/moxin/issues/178 hints at other problems with the event/hover handling in these widgets, perhaps there are relevant bugs in that code.

I'll update you with my findings, hopefully we can handle this within Makepad or the within the Widgets. 
(feel free to look into it as well if you want)
Thank you for this work, it was good starting point to find out the root problem.

joulei:
@ZhangHanDong update:
We've been discussing solving this issue within Makepad. 
We'll build some mechanism to properly update focus of underlying widgets whenever an Overlay takes over, is a bit tricky but it should be solved at the framework level so that we don't have to deal with it on the application side.
We can leave this on hold for a bit.

jmbejar:
@ZhangHanDong We have been working to solve this issue and similar ones. However, as @joulei commented above, the most appropriate solutions were changes or additions to Makepad widgets so we can better control cursor and hover states, specially around buttons. Also, we're reworking how Modal and other floating elements works in Moxin, so we can better deal with this kind of issues.

Although I'm closing this PR (since the issue is gone now) I want to say thanks, because your work and all the conversation with @joulei was very helpful to diagnose the problem and find the right solutions at the framework level 💪

PR #182, opened 2024-07-29
Fix downloading cancelling issues

### LINKED ISSUE #175: When deleting a model that is being downloaded, the interface displays incorrectly
https://github.com/user-attachments/assets/0058f837-c6ce-45a5-9f08-6518ca54a6e7

When deleting a model that is being downloaded, the interface displays incorrectly. This should be a bug.

jmbejar:
It fixes #175 

The cause of this issue is one limitation in our current backend implementation. If frontend sends a `PauseDownload` or `CancelDownload` command right after a `DownloadFile`, it doesn't have any effect. It takes some time (sometimes a few seconds) to be "listening" for those commands.

While we don't have a solution in the backend (where I think it would require a bit or refactor or deep thinking), this PR limits the visibility of Pause and Cancel buttons, making them invisible when the download is still being prepared by the backend.

A further improvement could be to add a loading spinner when buttons are not displayed, but it is not part of this PR.


jmbejar:
@L-jasmine It would be great to have your opinion here. As explained in the description, I noticed the backend can not process `PauseDownload` or `CancelDownload` commands after some time (sometime seconds) after receiving the `DownloadFile`. This is happening because, by design, the backend will create the channel to receive the "stop" messages right after receiving the first bit of data from the download link, which takes some time.

Do you think it is worth to tackle this issue in the backend side? I'd say yes, because it would allow user to cancel a download at any time, even right after starting one. It could be less confusing than the results I'm getting by just hiding the buttons in the UI.

Please let me know your thoughts, so I can decide if we should improve further what I'm doing in this PR or just wait for a backend solid solution.

noxware (approved):
lgtm

also did a quick test and seems to work as expected

PR #184, opened 2024-07-31
Use toggle panel from makepad

noxware:
Screenshot 2024-07-31 at 11 58 13


src/shared/toggle_panel.rs
```diff
@@ -5,116 +5,54 @@ live_design! {
     import makepad_widgets::theme_desktop_dark::*;
```
noxware, line 5: Please ignore this file update.

I can't import this widget from makepad right now so I'm updating the version here in Moxin.
We should delete this file and use makepad's exported panel instead before merging.

Cargo.toml
```diff
@@ -29,7 +27,7 @@ moxin-protocol = { path = "moxin-protocol" }
 moxin-backend = { path = "moxin-backend" }
 moxin-fake-backend = { path = "moxin-fake-backend" }
 
-makepad-widgets = { git = "https://github.com/jmbejar/makepad", branch = "moxin-release-v1" }
```
noxware, line 32: We need to revert this before merging.

But cherry picking the toggle panel commits on top of `moxin-release-v1` doesn't work. In that branch, the `import View` inside toggle panel causes a stack overflow (similar to what would happen in the `rik` branch if you try to import something from `base` inside toggle panel.

We may need to update moxin-release-v1 with stuff from rik.

Cargo.toml
```diff
@@ -29,7 +27,7 @@ moxin-protocol = { path = "moxin-protocol" }
...
+makepad-widgets = { git = "https://github.com/noxware/makepad", rev = "7ebad2443" }
```
noxware, line 30: This is a jmbejar's fork compatible branch with the widget added

https://github.com/noxware/makepad/tree/moxin-panel-for-jmbejar-remote

jmbejar: Done!

jmbejar (approved):

PR #191, opened 2024-08-03
fix issue 132

Guocork:


Guocork:
I tried to adjust the layout of <RotatedImage> using align, but it didn't work.

Guocork:
I tried to adjust the layout of <RotatedImage> using align, but it didn't work.

Guocork:
I tried to adjust the layout of RotatedImage using align, but it didn't work.

src/chat/model_selector_list.rs
```diff
@@ -134,7 +134,9 @@ impl Widget for ModelSelectorList {
         cx.begin_turtle(walk, self.layout);
 
         if self.visible {
-            self.draw_items(cx, &store.downloads.downloaded_files);
+            if let Some(loaded_file) = store.get_loaded_downloaded_file() {
```
jmbejar, line 137: Here is an issue. It could happen that there is no `loaded_file`, for this reason is an `Option` value.

You can replicate this corner case if you go to "My Models" and remove the current active model, and go back to the chat screen:

image

Looks like you can send the result of `store.get_loaded_downloaded_file()` and attempt to unwrap it in the `draw_items` function.

jmbejar (commented):
Thanks for all your work on this issue. The code looks good 👏 👏 
However, I found an issue to fix before merging it. Please let me know if you need extra help or information to wrap up the PR.

jmbejar (approved):
I found a minor issue with a loading bar in the model selector, but let's get this merged. I can fix it in a separate commit.

Thanks! 🎉

Guocork:
Anyway,I also found a situation where the arrow icon shouldn't appear when there are no available models, like this.
截图 2024-08-07 10-12-55

Guocork:
@jmbejar ,I think the arrow icon shouldn't appear when there are no available models, like this.
image
I don't know if this situation is the minor issue you mentioned.

PR #194, opened 2024-08-06
Feat/ctx size

L-jasmine:


moxin-protocol/src/protocol.rs
```diff
@@ -28,9 +28,12 @@ pub struct LoadModelOptions {
     pub prompt_template: Option<String>,
     pub gpu_layers: GPULayers,
     pub use_mlock: bool,
+    // if 0, the backend will use the default value
+    pub n_batch: u32,
```
jmbejar, line 32: What about using `Option` for those fields? Looks cleaner than relying on zero as the "null value".

L-jasmine: Update

jmbejar (approved):
Nice! :shipit:

PR #198, opened 2024-08-06
moxin-runner: support installing no-AVX versions of WasmEdge on Linux

kevinaboos:
Also, improve the error output for developers using moxin-runner to explicitly state when the main moxin binary hasn't yet been built before running moxin-runner.


kevinaboos:
tested working on all platforms.

jmbejar (approved):

PR #200, opened 2024-08-07
Windows cuda support

kevinaboos:
Depends on #198.

~~Things... sort of work? There seems to be a strange bug where the application hangs until the user presses a key in the terminal. But that only happens when the CUDA version of the WASI-nn plugin is active.~~
this is an unrelated bug.


kevinaboos:
ready to be merged.

jmbejar (approved):

PR #205, opened 2024-08-07
Use (or install) WasmEdge in the app-specific data directory *ONLY*

kevinaboos:
Relying on the system-wide installation has myriad problems; this avoids those many problems entirely by only checking the app's data directory for a WasmEdge installation and installing WasmEdge there if it is missing, which guarantees that the specific version and build characteristics of the WasmEdge installation is always under the sole control of Moxin.

As a side note, I'm not sure why there are so many tiny version changes to the `Cargo.lock` file. It seems like every time i build the Moxin workspace on a different host OS, there are tons of minor version changes. Shouldn't hurt anything, but it seems odd.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.21.0"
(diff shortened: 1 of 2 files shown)


kevinaboos:
tested and working on macOS.  Linux and Windows still pending.

kevinaboos:
tested working on Linux

kevinaboos:
ready to be merged

jmbejar (approved):

PR #210, opened 2024-08-08
Lazy model switch

noxware:
- [Design](https://www.figma.com/design/94gxq1QD6WTaNDJh1SGWsh/MoxinUI---New?node-id=102-2781&t=xNAlPIcDYiCtzYxR-0)


src/chat/chat_panel.rs
```diff
@@ -538,7 +541,11 @@ impl WidgetMatchEvent for ChatPanel {
                     self.redraw(cx);
                 }
                 ChatLineAction::Edit(id, updated, regenerate) => {
-                    store.chats.edit_chat_message(id, updated, regenerate);
+                    if regenerate {
```
noxware, line 544: The flag was completely changing what the function does, so I spitted the function (motivated by [this](https://martinfowler.com/bliki/FlagArgument.html)).

src/chat/model_selector.rs
```diff
@@ -320,55 +320,72 @@ impl ModelSelector {
...
+            file_name.to_string()
+        };
+
+        let is_loaded_file = match (file, loaded_file) {
+            (Some(f), Some(lf)) => f.id == lf.id,
+            (None, None) => true,
+            _ => false,
+        };
+
+        let text_enabled_color = hex_rgb_color(0x000000);
```
noxware, line 363: I don't know if makepad gives us some helper to build vec4 colors outside of the DSL so I made this util. 
This avoids needing to clarify which color this vector represents and also is recognized by vs code extensions like [colorize](https://marketplace.visualstudio.com/items?itemName=kamikillerto.vscode-colorize).

src/data/chats/chat.rs
```diff
@@ -268,51 +276,60 @@ impl Chat {
...
-                        }
 
-                        SignalToUI::set_ui_signal();
-                        if is_done {
+        let store_chat_tx = self.messages_update_sender.clone();
+        let wanted_file = wanted_file.clone();
+        let command_sender = backend.command_sender.clone();
+        thread::spawn(move || {
+            if let Err(err) = model_loader.load_blocking(wanted_file.id, command_sender.clone()) {
+                eprintln!("Error loading model: {}", err);
```
noxware, line 293: Future idea: Eventually, all `eprintln` we have around in the codebase could trigger a `Cx::post_action(ErrorNotification(...))` with the latest updates in makepad, to get feedback in the UI.

src/data/chats/model_loader.rs
```diff
@@ -1,88 +1,199 @@
...
+        Arc, Mutex,
+    },
+    thread,
+};
+
+/// Immutable handleable error type for the model loader.
+///
+/// Actually, just a wrapper around `anyhow::Error` to support `Clone`.
+#[derive(Debug, Clone)]
+pub struct ModelLoaderError(Arc<anyhow::Error>);
```
noxware, line 19: Anyhow error is not clone. If we want to send an error as a result and save it in the `Failed` status at the same time, we need to wrap it. And since this is pretty generic it could be a `MoxinError` actually.

But this is unnecessary, as we are not reading the error from the status currently. We are just reading the error in the result of `load_blocking`. So we actually don't need to be clone. So we can just remove this.

Unless you believe this can be useful. If not I will remove the error from `Failed` and this wrapper with it.

src/data/chats/model_loader.rs
```diff
@@ -1,88 +1,199 @@
...
+/// Unit for handling the non-blocking loading of models across threads.
+#[derive(Clone, Default)]
+pub struct ModelLoader(Arc<Mutex<ModelLoaderInner>>);
+
+impl ModelLoader {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn load_blocking(
```
noxware, line 72: Having blocking functions by default instead of implicit threading functions makes code more composable and avoids creating extra threads when you are already in a separate synchronous thread (like the thread spawned by send message function).

src/data/chats/model_loader.rs
```diff
@@ -1,88 +1,199 @@
...
         };
 
-        Ok(())
+        SignalToUI::set_ui_signal();
+        result
     }
-}
\ No newline at end of file
+
+    pub fn load(&mut self, file_id: FileID, command_sender: Sender<Command>) {
```
noxware, line 124: Provided just to have a function with the previous behavior although, forcing the caller yo be explicit about spawning a thread would be also an option.

src/data/chats/model_loader.rs
```diff
@@ -1,88 +1,199 @@
...
+
+impl ModelLoader {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn load_blocking(
+        &mut self,
+        file_id: FileID,
+        command_sender: Sender<Command>,
```
noxware, line 75: The backend type is just this actually and this works better in multithreading contexts as you can clone it. We could make the `Backend` type `Clone` just to have the pretty type `backend: Backend` and also remove the `Rc<Backend>` we have in store but is not a big deal for me.

src/chat/model_selector.rs
```diff
@@ -320,55 +320,72 @@ impl ModelSelector {
     }
 
     fn update_selected_model_info(&mut self, cx: &mut Cx, store: &Store) {
+        let is_loading = store.chats.model_loader.is_loading();
```
jmbejar, line 323: I'm personally prefer how the code was organized before these changes, because it is easier to spot how the UI looks in each case.

Since this refactor was not related with this PR, can we just revert it?

src/data/chats/mod.rs
```diff
@@ -192,7 +134,7 @@ impl Chats {
         self.saved_chats.push(new_chat);
     }
 
-    pub fn create_empty_chat_and_load_file(&mut self, file: &File) {
```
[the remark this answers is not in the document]
noxware: Probably a leftover. I will revert these changes 👍

src/chat/model_selector.rs
```diff
@@ -449,6 +466,17 @@ fn options_to_display(store: &Store) -> bool {
 }
 
 fn no_active_model(store: &Store) -> bool {
-    store.get_loaded_downloaded_file().is_none()
-        && store.chats.get_currently_loading_model().is_none()
+    store.get_loaded_downloaded_file().is_none() && store.get_loading_file().is_none()
+}
+
+fn rgb_color(r: u8, g: u8, b: u8) -> Vec4 {
```
fmzbl, line 472: why is this another function? its only used on `hex_rgb_color`, why not just doing:
```rust
fn hex_rgb_color(hex: u32) -> Vec4 {
    vec4(
        ((hex >> 16) & 0xFF) as f32 / 255.0, // r
        ((hex >> 8) & 0xFF) as f32 / 255.0,  // g
        (hex & 0xFF) as f32 / 255.0,         // b
        1.0,                                 // a
    )
}
```

fmzbl: also, i would move to utils or somthing like that

jmbejar (approved):
Good job!

PR #211, opened 2024-08-09
fix-issue-208

### LINKED ISSUE #208: "Searching..." overlay does not automatically disappear once the search is complete
After searching for a term on the "Discover" screen, the `Searching...` animation will be shown forever until the user scrolls on the screen. 

See screenshot below, which will remain on the screen indefinitely until the user scrolls up or down.

I assume the problem is related to something like a forgotten call to `redraw()` after the search results have been fully returned.

image

Guocork:
Fixes #208 

I add the function `redraw()`,and it seems work.But the modal closes very slowly. I see that the is_loading variable in the `update_loading_and_error_message()` function changes to true first during the search process, then to false, and only then does the modal close. It might be that my internet speed is slow.
[Here is the specific location in the code.](https://github.com/moxin-org/moxin/blob/main/src/landing/model_list.rs#L156)

### DIFF
--- a/src/landing/model_list.rs
+++ b/src/landing/model_list.rs
@@ -83,6 +83,7 @@ impl Widget for ModelList {
         if self.loading_delay.is_event(event).is_some() {
             self.update_loading_and_error_message(cx, scope);
         }
+        self.redraw(cx);
     }
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {


src/landing/model_list.rs
```diff
@@ -83,6 +83,7 @@ impl Widget for ModelList {
         if self.loading_delay.is_event(event).is_some() {
             self.update_loading_and_error_message(cx, scope);
         }
+        self.redraw(cx);
```
jmbejar, line 86: Asking the framework to redraw the widget when *any* event happens could lead poor performance, just because ir could be redrawing most of the time (for instance, any mouse move event will trigger a redraw here). Ideally, we should invoke `self.redraw(cx)` when certain condition is met.

jmbejar:
> I add the function `redraw()`,and it seems work.But the modal closes very slowly. I see that the is_loading variable in the `update_loading_and_error_message()` function changes to true first during the search process, then to false, and only then does the modal close. It might be that my internet speed is slow. [Here is the specific location in the code.](https://github.com/moxin-org/moxin/blob/main/src/landing/model_list.rs#L156)

I was not able to reproduce the original problem, so I can not easily say where we could have a missing call to `redraw`. I will try to investigate a bit more on Monday.

src/landing/model_list.rs
Guocork: @jmbejar Thanks for your reviewing. Also, I have a question about issue 132, it has already been merged. I think the arrow icon shouldn't appear when there are no available models, like this.
image
I don't know if this situation is the minor issue you mentioned.

jmbejar: > @jmbejar Thanks for your reviewing. Also, I have a question about issue 132, it has already been merged. I think the arrow icon shouldn't appear when there are no available models

This issue is something I was not aware... so feel free to drop a PR to fix this problem, if you want 👍

Guocork: I would be more than happy to fix this problem. I will try to solve it.

joulei:
We've been discussing this offline and will review the issue within Makepad. 
There's either some missing redrawing in the Linux platform, or there's really a missing redraw in the application however it's not problematic in macOS because that platform is doing extra redrawing.

joulei:
Closing, will handle this issue at the framework level.

Issue #225, opened 2024-08-20
WasmEdge CUDA detection not working correctly


kevinaboos:
Line 36 of the WasmEdge [`install_v2.sh` script](https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install_v2.sh) attempts to discover the CUDA installation using `nvcc --version`, but it is hardcoded to look for `nvcc` in `/usr/local/cuda/bin/`.

On some machines, cuda is found in `/opt/...`, so this will fail.

@juntao Line 36 of the install script should use `nvcc` from the current path first, and then fall back to `/usr/local/cuda/bin/nvcc`, and then fall back to `/opt/cuda/bin/nvcc`, and finally consider cuda as non-existent.

`moxin-runner` already does this check locally in Rust code, so a short-term fix would be to directly pass in the cuda version into the invocation of `install_v2.sh`, i.e., `-c 11` or `-c 12`.

kevinaboos:
Another potentially related issue is that CUDA doesn't actually get used even when WasmEdge successfully detects the CUDA installation.

Here are some examples of warnings/errors I observe on my Linux and Windows machines that have CUDA 12.5, but with an older Nvidia GeForce 1070Ti graphics card:

```
[warning] [WASI-NN] llama.cpp: ggml_backend_cuda_graph_compute: disabling CUDA graphs due to GPU architecture
```
```
[error] [WASI-NN] llama.cpp: ggml_cuda_init: failed to initialize CUDA: CUDA driver version is insufficient for CUDA runtime version
```
```
[warning] [WASI-NN] llama.cpp: ggml_cuda_host_malloc: failed to allocate 384.00 MiB of pinned memory: CUDA driver version is insufficient for CUDA runtime version
```

kevinaboos:
> `moxin-runner` already does this check locally in Rust code, so a short-term fix would be to directly pass in the cuda version into the invocation of `install_v2.sh`, i.e., `-c 11` or `-c 12`.

this was implemented in #226

L-jasmine:
@hydai

kevinaboos:
Closed by https://github.com/WasmEdge/WasmEdge/pull/3682

thanks for the fix @hydai and @L-jasmine. much appreciated!

PR #226, opened 2024-08-20
Manually detect and force CUDA version passed into `install_v2.sh`

### LINKED ISSUE #207: `moxin-runner`: de-dup/clarify how environment variables are set
There are currently two functions, one for Unix and one for Windows, that handle how environment variables are set. We should simplify things into a single function, which would help ensure that we set all env vars _consistently_, which is currently not happening.

Also, we could consider adding a mode where we don't use [`std::env::set_var()`](https://doc.rust-lang.org/std/env/fn.set_var.html) to set the current process's environment variables, which is debatably unsafe/dangerous according to the documentation of that function.
Instead, we could just apply those env vars to the `std::process::Command(...)` invocation that spawns the main moxin binary as a child process. That way, the env vars for the current process are not modified, but rather just set for the child process that runs the main Moxin app.

---------------
Separately, when running `cargo run -p moxin-runner -- --install`, we should emit instructions that inform developers how to set the necessary env variables based on the WasmEdge components that we just installed.
* `WASMEDGE_DIR`
* `WASMEDGE_PLUGIN_PATH` (may not be required on all platforms)?
* `PATH` (prepend `WASMEDGE_DIR/bin` to `PATH`)

kevinaboos:
The `install_v2.sh` script does not properly detect CUDA installations at the moment (see #255), so we do the detection ourselves in `moxin-runner` and then pass in the correctly-detected CUDA version as an argument into that bash script.

Partially addresses #225.

Also, this PR allows `moxin-runner` to behave as a general cargo command runner:
```sh
cargo run -p moxin-runner -- cargo build --release
```
which saves Moxin devs a lot of time & effort because they don't have to manually install WasmEdge and set up all the necessary environment variables. Let `moxin-runner` do it for you!

Also closes #207.

### DIFF
--- a/moxin-runner/src/main.rs
+++ b/moxin-runner/src/main.rs
@@ -69,6 +69,7 @@
 #![allow(unused)]
 
 use std::{
+    ffi::OsStr,
     path::{Path, PathBuf},
     process::Command,
 };
@@ -223,6 +224,7 @@ fn main() -> std::io::Result<()> {
 
 #[cfg(not(feature = "macos_bundle"))]
 fn main() -> std::io::Result<()> {
+
     assert_cpu_features();
 
     let (wasmedge_root_dir_in_use, main_dylib_path, wasi_nn_plugin_path) = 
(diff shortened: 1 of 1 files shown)


kevinaboos:
Still needs some testing, marking as draft until I can test it thoroughly.

jmbejar (approved):

PR #234, opened 2024-08-23
Agents UI

noxware:
# Description

First UI prototype over mae integration.

# Screenshot

Screenshot 2024-08-23 at 20 52 24


resources/images/agent.png
noxware, line 1: Custom made image, so don't worry.

src/chat/agent_button.rs
noxware, line 1: Used in both, left panel portal list and prompt input `@` "autocomplete".

src/chat/chat_panel.rs
```diff
@@ -185,74 +183,7 @@ live_design! {
         }
     }
 
-    PromptButton = <CircleButton> {
```
noxware, line 188: Most of this has been extracted to `prompt_input.rs`.

src/chat/chat_panel.rs
```diff
@@ -764,10 +710,17 @@ impl ChatPanel {
...
             let store = scope.data.get_mut::<Store>().unwrap();
-            //store.send_chat_message(prompt.clone());
 
-            // MAE
-            store.send_agent_message(MaeAgent::Questioner, prompt.clone());
+            if let Some(agent_selected) = self
+                .prompt_input(id!(main_prompt_input))
+                .borrow()
+                .unwrap()
+                .agent_selected
```
noxware, line 707: Made the selected agent public so the send logic remains here but probably it should be moved inside the prompt input.

src/chat/chat_panel.rs
noxware: Although, logic controlling most of this remains here. Probably should be abstracted inside.

PR #240, opened 2024-08-31
Make models search case insensitive

jmbejar:


moxin-backend/src/store/model_cards.rs
```diff
@@ -461,15 +461,16 @@ impl ModelCardManager {
         limit: usize,
         offset: usize,
     ) -> reqwest::Result<Vec<ModelIndex>> {
+        let search_text = search_text.trim().to_ascii_lowercase();
```
noxware, line 464: Bonus: A cheap way to improve this search would be to replace `.trim()` with `.split_whitespace().collect::<Vec<_>>().join(" ")`. This will do the same as `trim` but also will get rid of spaces between search terms.

Comparison:

```rust
let s = "    aaaa    bbbb    cccc  ".trim();
println!("{:?}", s);

let s = "    aaaa    bbbb  cccc  ".split_whitespace().collect::<Vec<_>>().join(" ");
println!("{:?}", s);
```

```rust
"aaaa    bbbb    cccc"
"aaaa bbbb cccc"
```

noxware: We could also omit `.collect::<Vec<_>>().join(" ")` and simply check each word independently to perform order-unaware search where `yi-1.5 9b` and `9b yi-1.5` would match the same results. 

But that would be a different kind of search.

noxware (approved):
Works 👍 Suggested an easy improvement as an extra.

PR #245, opened 2024-09-05
Save the widows size users set

Guocork:


joulei:
Closing this due to inactivity, we'll likely implement this similarly to how Robrix has done it recently. 
Thanks!

PR #246, opened 2024-09-06
fix-issue-206

Guocork:
I solved the issue by removing the model-generated reply from the corresponding array when the user clicks the cancel button. However, this creates a situation where the user's questions may appear next to each other. I'm not sure if this is the best solution.
截图 2024-09-06 11-25-59

### DIFF
--- a/src/data/chats/mod.rs
+++ b/src/data/chats/mod.rs
@@ -108,7 +108,10 @@ impl Chats {
         if let Some(chat) = self.get_current_chat() {
             chat.borrow_mut().cancel_streaming(self.backend.as_ref());
             let mut chat = self.get_current_chat().unwrap().borrow_mut();
-            chat.messages.pop();
+            let message = chat.messages.last_mut().unwrap();
+            if message.content.trim().is_empty() {
+                chat.messages.pop();
+            }
         }
     }
 
--- a/src/data/chats/mod.rs
+++ b/src/data/chats/mod.rs
@@ -107,6 +107,8 @@ impl Chats {
     pub fn cancel_chat_streaming(&mut self) {
         if let Some(chat) = self.get_current_chat() {
             chat.borrow_mut().cancel_streaming(self.backend.as_ref());
+            let mut chat = self.get_current_chat().unwrap().borrow_mut();
+            chat.messages.pop();
         }
     }
 


jmbejar:
@Guocork Thanks for working on this issue.

I see a big problem with this approach: the cancel button is mostly used to interrupt the model streaming of the response, allowing to keep part of the generated output. It is true that the response should be deleted if the streaming has not started yet.

I have been thinking on this problem but in a separate [pull request](https://github.com/moxin-org/moxin/pull/218) where we are integrating with another backend for agents.

Since there is no immediate plans to integrate all this PR, maybe you want to help to port some changes in the chat state management from this PR to this one?

If you're interested, please look at how in this PR there is a change in the `Chat` struct from having a bool `is_streaming` to `state` (which is an enum with a representation from the `cancelled` state among others).
https://github.com/moxin-org/moxin/pull/218/files#diff-cce49c6f311f98aee6e76e747d766858047b8a1711514e95ba9fc5344daa1eabL92-R120
Once the chat has this new `state` field, it would be possible to model the cancelled state, specially before the initial chunk of the response comes from the backend model.

Please let me know if you need more guidance or ideas.

Guocork:
> @Guocork Thanks for working on this issue.
> 
> I see a big problem with this approach: the cancel button is mostly used to interrupt the model streaming of the response, allowing to keep part of the generated output. It is true that the response should be deleted if the streaming has not started yet.
> 
> I have been thinking on this problem but in a separate [pull request](https://github.com/moxin-org/moxin/pull/218) where we are integrating with another backend for agents.
> 
> Since there is no immediate plans to integrate all this PR, maybe you want to help to port some changes in the chat state management from this PR to this one?
> 
> If you're interested, please look at how in this PR there is a change in the `Chat` struct from having a bool `is_streaming` to `state` (which is an enum with a representation from the `cancelled` state among others). https://github.com/moxin-org/moxin/pull/218/files#diff-cce49c6f311f98aee6e76e747d766858047b8a1711514e95ba9fc5344daa1eabL92-R120 Once the chat has this new `state` field, it would be possible to model the cancelled state, specially before the initial chunk of the response comes from the backend model.
> 
> Please let me know if you need more guidance or ideas.

OK，I got it. i will try to fix that.

Guocork:
I added a piece of logic, and it seems to have achieved the desired effect.

jmbejar (approved):

PR #249, opened 2024-09-12
fix-issue-248

Guocork:


jmbejar (approved):
Tested it and seems to work OK 👏

PR #251, opened 2024-09-14
Add Clear Text Button to Models Search Bar

### LINKED ISSUE #237: Add a X button to clear the text input field in the model search screen
image

Should be a X to clear the input field. Currently only way to get back to the main discover page is to manually deleting all characters in the input field.

Guocork:
Fixes #237 

I want to implement a feature where the X button appears when the mouse clicks on the input field or when the user types something, and it stays hidden at other times. But I haven't found a good way to achieve this yet.

### DIFF
--- a/src/landing/search_bar.rs
+++ b/src/landing/search_bar.rs
@@ -14,6 +14,7 @@ live_design! {
     import crate::landing::sorting::Sorting;
 
     ICON_SEARCH = dep("crate://self/resources/icons/search.svg")
+    ICON_CLOSE = dep("crate://self/resources/icons/close.svg")
 
     SearchBar = {{SearchBar}} {
         width: Fill,
@@ -93,6 +94,17 @@ live_design! {
                 width: Fill,
                 height: Fit,
                 empty_message: "Search Model by Keyword"
+            }
+
+            x_button = <MoxinButton> {
+                visible: false,
+                draw_icon: {
+                    svg_file: (ICON_CLOSE),
+                    fn get_color(self) -> vec4 {
+                        return #666;
+                    }
+                }
+                icon_walk: {width: 17, height: 17}
             }
         }
 
@@ -179,6 +191,7 @@ impl Widget for SearchBar {
 impl WidgetMatchEvent for SearchBar {
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
         let input = self.text_input(id!(input));
+        let x_button = self.button(id!(x_button));
 
         if let Some(keywords) = input.returned(actions) {
             if keywords.len() > 0 {
@@ -195,8 +208,13 @@ impl WidgetMatchEvent for SearchBar {
         }
 
         if let Some(_) = input.changed(actions) {
+            x_button.set_visible(true);
             cx.stop_timer(self.search_timer);
             self.search_timer = cx.start_timeout(self.search_debounce_time);
+        }
+
+        if self.button(id!(x_button)).clicked(actions) {
+            input.set_text_and_redraw(cx, "");
         }
     }
 }
--- a/src/landing/search_bar.rs
+++ b/src/landing/search_bar.rs
@@ -101,10 +101,10 @@ live_design! {
                 draw_icon: {
                     svg_file: (ICON_CLOSE),
                     fn get_color(self) -> vec4 {
-                        return #666;
+                        return #8;
                     }
                 }
-                icon_walk: {width: 17, height: 17}
+                icon_walk: {width: 10, height: 10}
             }
         }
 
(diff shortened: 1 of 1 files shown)


Guocork (commented):
Right now, after the user types something for the first time, the X button stays visible forever.

src/landing/search_bar.rs
```diff
@@ -195,9 +208,14 @@ impl WidgetMatchEvent for SearchBar {
         }
 
         if let Some(_) = input.changed(actions) {
+            x_button.set_visible(true);
```
Guocork, line 211: Right now, after the user types something for the first time, the X button stays visible forever.

joulei: The `changed` method returns an Option<String> with the contents of the `TextInput`, you can check if the value is empty to hide the button, something like this:
```rs
if let Some(value) = input.changed(actions) {
     if value.is_empty() {
          x_button.set_visible(false);
     } else {
         x_button.set_visible(true);
    }
    ...
}
```

Also, I think we should me the icon a bit smaller and lighter color, right now it draws too much attention.

Guocork: I fix that case by adding that logic. Thank you very much.

> The `changed` method returns an Option with the contents of the `TextInput`, you can check if the value is empty to hide the button, something like this:
> 
> ```rust
> if let Some(value) = input.changed(actions) {
>      if value.is_empty() {
>           x_button.set_visible(false);
>      } else {
>          x_button.set_visible(true);
>     }
>     ...
> }
> ```
> 
> Also, I think we should me the icon a bit smaller and lighter color, right now it draws too much attention.

src/landing/search_bar.rs
```diff
@@ -194,9 +207,21 @@ impl WidgetMatchEvent for SearchBar {
...
+            if val.is_empty() {
+                x_button.set_visible(false);
+                cx.stop_timer(self.search_timer);
+                self.search_timer = cx.start_timeout(self.search_debounce_time);
+            } else {
+                x_button.set_visible(true);
+                cx.stop_timer(self.search_timer);
+                self.search_timer = cx.start_timeout(self.search_debounce_time);
+            }
+        }
```
joulei, line 220: We can avoid duplicating code here, you can write something like this:
```rs
        if let Some(val) = input.changed(actions) {
            if val.is_empty() {
                x_button.set_visible(false);
            } else {
                x_button.set_visible(true);
            }
            cx.stop_timer(self.search_timer); // don't write and the below these lines twice
            self.search_timer = cx.start_timeout(self.search_debounce_time);
        }
 ```
I suggest to even shorten it further:
 ```rs
 if let Some(text) = input.changed(actions) {
    x_button.set_visible(!text.is_empty());
    cx.stop_timer(self.search_timer);
    self.search_timer = cx.start_timeout(self.search_debounce_time);
}
 ```

File: src/landing/search_bar.rs [hunk 3]
```diff
@@ -94,6 +95,17 @@ live_design! {
                 height: Fit,
                 empty_message: "Search Model by Keyword"
             }
+
+            x_button = <MoxinButton> {
```
joulei, line 99: Another suggestion, we should aim for semantic names that are also consistent with the naming of other elements in the UI.
If we name the button to something like `clear_input_button` or `clear_text_button`, what the element does becomes more obvious.

joulei (commented):
Added some comments about code quality, otherwise the logic looks good.

As a final note, some feedback about git and GitHub usage to keep in mind next time:
We should make the pull requests be consistent with others one in the repo. Usually the convention is to give the PR a useful name about what it does, and simply link the issue instead of using that as title, same for commits.

E.g.: 
- PR title: Add Clear Text Button to Models Search Bar
- Description: you can reference the issue using "Addresses #<issue-number>" or "Fixes #<issue-number>", and then mention anything you think is relevant.
- Commits: these are a bit more personal, but ideally they should tell what they changed, in this case the commits could've said:
    1. 88568de - Add clear text button to search bar
    2. 7c1cb84 - Hide clear button when input is empty
    
This way the pull requests become easier to identify and the reviews easier to navigate.

src/landing/search_bar.rs
Guocork: ok，I have improved the code. Thank you very much for your review!

src/landing/search_bar.rs
Guocork: thank you for your suggestion

Guocork:
> Added some comments about code quality, otherwise the logic looks good.
> 
> As a final note, some feedback about git and GitHub usage to keep in mind next time: We should make the pull requests be consistent with others one in the repo. Usually the convention is to give the PR a useful name about what it does, and simply link the issue instead of using that as title, same for commits.
> 
> E.g.:
> 
>     * PR title: Add Clear Text Button to Models Search Bar
> 
>     * Description: you can reference the issue using "Addresses #" or "Fixes #", and then mention anything you think is relevant.
> 
>     * Commits: these are a bit more personal, but ideally they should tell what they changed, in this case the commits could've said:
>       
>       1. [88568de](https://github.com/moxin-org/moxin/commit/88568de6159d04008ecf84e9f58654dd3015603d) - Add clear text button to search bar
>       2. [7c1cb84](https://github.com/moxin-org/moxin/commit/7c1cb8441a81027f3ea614e718f555a8a6225037) - Hide clear button when input is empty
> 
> 
> This way the pull requests become easier to identify and the reviews easier to navigate.

Thank you for your suggestions. I will keep them in mind for future code submissions.

joulei (approved):
@Guocork good work! thanks for going over the suggestions.

PR #253, opened 2024-09-18
[Backend] fix model name

L-jasmine:
* Fixed chat model name as "moxin-chat"
* Fixed embedding model name as "moxin-embedding"

The frontend doesn't need to change because I've already replaced the model name in the request on the backend


juntao:
Thanks. Would love to see this merged. 

This is needed for agent apps that use the moxin API server. We cannot ask the agent to re-configure its model name whenever the moxin UI user switch to another model to chat.

jmbejar (approved):

PR #258, opened 2024-09-20
Add rust playground link

Guocork:
On hold until we have the ability to wrap the code blocks within the Markdown element, in separate views (where we can place actions). Currently this PR adds the link at the end of the message and only works for the first code block.


joulei:
Hi @Guocork, @jmbejar mentioned that the link my not be too obvious of what is doing, and maybe placement is not the best.

@Guocork the icon you're using is used for sharing content, it usually opens an option panel or copies a link.

Instead, this icon is used more commonly for external links (navigates outside of the current context). We should either use something like this or a rust logo.
image

And to make it clear, we can use a tooltip on hover (similar to the tooltips used in the right panel) with a message like "Open in Rust Playground". You can check how tooltips are used in `src/chat/chat_params.rs`.

Besides that, one current limitation (in this PR) is that this would copy the first code block in a markdown message. If the message has multiple code blocks, the rest are ignored.
The only solution to this I think is to have the link next to each code block, similar to what other platforms do:

image

However this would require writing our own custom widget similar to the original `Markdown` widget, with the added logic to render the code blocks differently with an added actions section to it. This seems like a complex task.

@jmbejar should we leave this PR on hold until we have support for that?

jmbejar:
> However this would require writing our own custom widget similar to the original `Markdown` widget, with the added logic to render the code blocks differently with an added actions section to it. This seems like a complex task.
> 
> @jmbejar should we leave this PR on hold until we have support for that?

Correct, we don't have any related ability in the `Markdown` widget yet, so it is hard to implement this feature in a good-looking way. 
Let me check how hard is to have something in the markdown widget to properly cover this case. In the meantime, let's hold this PR but the idea will be to give @Guocork the tools to complete it 👍

joulei:
Closing as we have now replaced the chat with a MolyKit implementation, and I don't we'll plan to implement this in the near future.

PR #260, opened 2024-09-23
Fix visibility of markdown-formatted message

jmbejar:
This PR bumps the Makepad version. Some adjustments were needed to avoid a visual problem that was very noticeable in markdown-formatted messages.

This screenshot demonstrates the font color issue (bumping to most recent Makepad version, without the other changes in this PR)
image


noxware (approved):
lgtm

PR #262, opened 2024-09-27
Add a dot at the end of chat lines during model replies to make the responses look smoother.

Guocork:
There is a problem: Sometimes Makepad cannot immediately render this dot.

### DIFF
--- a/src/chat/chat_line.rs
+++ b/src/chat/chat_line.rs
@@ -420,19 +420,19 @@ impl ChatLineRef {
             return;
         };
 
-        if is_streaming && !text.is_empty() {
-            let output = format!("{}{}", text, "●");
-            inner.text_input(id!(input)).set_text(&output.trim());
-            inner.label(id!(plain_text_message)).set_text(&output.trim());
-            inner.markdown(id!(markdown_message)).set_text(&output.trim());
-        } else {
-            inner.text_input(id!(input)).set_text(text.trim());
-            inner.label(id!(plain_text_message)).set_text(text.trim());
-            inner.markdown(id!(markdown_message)).set_text(text.trim());
-        }
         match inner.edition_state {
-
+            
             ChatLineState::Editable | ChatLineState::NotEditable => {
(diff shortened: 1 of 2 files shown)


src/chat/chat_line.rs
```diff
@@ -415,16 +415,24 @@ impl ChatLineRef {
...
+        if is_streaming && !text.is_empty() {
+            let output = format!("{}{}", text, "●");
+            inner.text_input(id!(input)).set_text(&output.trim());
+            inner.label(id!(plain_text_message)).set_text(&output.trim());
+            inner.markdown(id!(markdown_message)).set_text(&output.trim());
+        } else {
+            inner.text_input(id!(input)).set_text(text.trim());
+            inner.label(id!(plain_text_message)).set_text(text.trim());
+            inner.markdown(id!(markdown_message)).set_text(text.trim());
+        }
```
jmbejar, line 432: We're introducing an issue here. The user is not longer able to edit a message. It's happening because you moved the `inner.text_input(id!(input)).set_text(..)` out of the `ChatLineState::Editable | ChatLineState::NotEditable` arms of the `match inner.edition_state`. 

I feel that a better way to approach this change is to compute the change to the `text` (adding the ●) before calling this function. The caller has everything in place to do it.

jmbejar (changes requested):

src/chat/chat_line.rs
Guocork: I moved the logic check back into the `match` , and it seems to have restored the previous effect. I tried adding the  ● in the `chat_panel`, but the code isn’t as clean with this approach.

joulei:
@Guocork thanks for the PR
I'm investigating why Makepad cannot sometimes can't render the character

joulei:
@jmbejar we confirmed that the rendering issue with the special character only happens in Linux.
Are you ok with merging this as-is, while we open the issue in Makepad?

joulei (approved):

PR #267, opened 2024-09-30
Override the original font

Guocork:
I tried overriding the font style in markdown, and it seems to have worked.
截图 2024-09-30 14-56-46


joulei:
@Guocork thanks for working on this, however @jmbejar hadn't seen that you were on it so his PR has already fixed the issue. There's a slight difference in that the final decision was to use Makepad fonts only.
No other changes are needed.

PR #279, opened 2024-10-11
Mofa integration

jmbejar:


noxware:
- Mofa feature hidden by default.
- Use `MOFA_FRONTEND=visible` to show it.
- Use `MOFA_FRONTEND=visible MOFA_BACKEND=real` to show it and use the dora implementation instead of the fake one.

joulei (approved):
I think this has reached a stable-enough state where it's safe to merge, at least into `dev`.
There might be some edge cases around agents or minor bugs elsewhere, but we probably won't catch them until we start to work daily on top of these changes.

PR #295, opened 2024-11-04
Signals to explicit actions

noxware:
# Changes
- Replaced `Event::Signal` pattern matchings with reactions to specific cross-thread actions.
- Replaced `SignalToUI::set_ui_signal()` with `Cx::post_action(...)`.
- Replaced "polling" in the store when a signal comes, with normal action handling and routing.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4,7 +4,8 @@ use crate::data::store::*;
 use crate::landing::model_files_item::ModelFileItemAction;
 use crate::shared::actions::{ChatAction, DownloadAction};
 use crate::shared::download_notification_popup::{
-    DownloadNotificationPopupAction, DownloadNotificationPopupWidgetRefExt, DownloadResult, PopupAction
+    DownloadNotificationPopupAction, DownloadNotificationPopupWidgetRefExt, DownloadResult,
+    PopupAction,
 };
 use crate::shared::popup_notification::PopupNotificationWidgetRefExt;
 use makepad_widgets::*;
@@ -189,6 +190,8 @@ impl MatchEvent for App {
             );
 
         for action in actions.iter() {
+            self.store.handle_action(action);
+
             match action.cast() {
                 StoreAction::Search(keywords) => {
                     self.store.search.load_search_results(keywords);
@@ -248,7 +251,9 @@ impl MatchEvent for App {
                 DownloadNotificationPopupAction::ActionLinkClicked
                     | DownloadNotificationPopupAction::CloseButtonClicked
             ) {
-                self.ui.popup_notification(id!(popup_notification)).close(cx);
+                self.ui
+                    .popup_notification(id!(popup_notification))
+                    .close(cx);
             }
         }
     }
--- a/src/app.rs
+++ b/src/app.rs
@@ -155,11 +155,6 @@ impl LiveRegister for App {
 
 impl AppMain for App {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event) {
-        // Process all possible store incoming events
-        if let Event::Signal = event {
-            self.ui.redraw(cx);
-        }
-
         let scope = &mut Scope::with_data(&mut self.store);
         self.ui.handle_event(cx, event, scope);
         self.match_event(cx, event);
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,4 +1,5 @@
 use crate::chat::chat_panel::ChatPanelAction;
+use crate::data::downloads::download::DownloadFileAction;
 use crate::data::downloads::DownloadPendingNotification;
 use crate::data::store::*;
 use crate::landing::model_files_item::ModelFileItemAction;
@@ -185,9 +186,7 @@ impl MatchEvent for App {
         for action in actions.iter() {
             self.store.handle_action(action);
 
-            if let Some(_) =
-                action.downcast_ref::<crate::data::downloads::download::DownloadFileAction>()
-            {
+            if let Some(_) = action.downcast_ref::<DownloadFileAction>() {
                 self.notify_downloaded_files(cx);
             }
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -157,7 +157,6 @@ impl AppMain for App {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event) {
         // Process all possible store incoming events
         if let Event::Signal = event {
-            self.notify_downloaded_files(cx);
             self.ui.redraw(cx);
         }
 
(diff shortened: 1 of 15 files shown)


src/app.rs
```diff
@@ -189,6 +184,12 @@ impl MatchEvent for App {
             );
 
         for action in actions.iter() {
+            self.store.handle_action(action);
+
+            if let Some(_) = action.downcast_ref::<DownloadFileAction>() {
+                self.notify_downloaded_files(cx);
+            }
+
```
noxware, line 192: These replace the `if Event::Signal = event { ... }` handling that was in this file.

src/chat/chat_history.rs
```diff
@@ -73,13 +73,6 @@ impl Widget for ChatHistory {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
         self.deref.handle_event(cx, event, scope);
         self.widget_match_event(cx, event, scope);
-
-        // TODO This is a hack to redraw the chat history and reflect the
-        // name change on the first message sent.
-        // Maybe we should send and receive an action here?
-        if let Event::Signal = event {
-            self.redraw(cx);
-        }
```
noxware, line 82: Note: Idk why, but removing this didn't cause any issues. Am I missing something?

src/data/chats/chat.rs
```diff
@@ -16,8 +16,14 @@ use super::model_loader::ModelLoader;
 
 pub type ChatID = u128;
 
-#[derive(Clone, Debug)]
-pub enum ChatTokenArrivalAction {
+#[derive(Debug)]
+pub struct ChatEntityAction {
+    pub chat_id: ChatID,
```
noxware, line 21: This and the `DownloadFileAction` now have an `id` so the actions can be routed back to the entity that should handle the message. 

These ids are also used on widgets to detect if they should redraw or not (compared to the previous signal based approach that would redraw everything).

src/data/chats/chat.rs
```diff
@@ -1,11 +1,11 @@
 use anyhow::{anyhow, Result};
-use makepad_widgets::SignalToUI;
+use makepad_widgets::Cx;
 use moly_backend::Backend;
 use moly_protocol::data::{File, FileID};
 use moly_protocol::open_ai::*;
 use moly_protocol::protocol::Command;
 use std::path::PathBuf;
-use std::sync::mpsc::{channel, Receiver, Sender};
+use std::sync::mpsc::channel;
```
noxware, line 8: We still need a channel but only to communicate with the backend.

src/data/chats/mod.rs
```diff
@@ -192,4 +196,14 @@ impl Chats {
             }
         }
     }
+
+    pub fn handle_action(&mut self, action: &Box<dyn ActionTrait>) {
```
noxware, line 200: Refine and forward the action into the specific chat that needs to handle it.

src/data/chats/model_loader.rs
```diff
@@ -84,7 +87,7 @@ impl ModelLoader {
             Err(anyhow!("Internal communication error"))
         };
 
-        SignalToUI::set_ui_signal();
+        Cx::post_action(ModelLoaderAction);
```
noxware, line 90: No much to notify here :P Model loader is pretty automatic.

src/data/store.rs
```diff
@@ -258,36 +260,21 @@ impl Store {
...
         Ok(())
     }
 
-    pub fn process_event_signal(&mut self) {
-        self.update_downloads();
-        self.update_chat_messages();
-        self.update_search_results();
-        self.update_load_model();
-    }
+    pub fn handle_action(&mut self, action: &Action) {
```
noxware, line 266: This replaced `process_event_signal` in spirit.

Instead of polling many data sources, this forwards the action (message) down into other modules so they decide if they need to do something.

src/data/store.rs
```diff
@@ -258,36 +260,21 @@ impl Store {
...
         }
-    }
 
-    fn update_chat_messages(&mut self) {
-        let Some(chat) = self.chats.get_current_chat() else {
-            return;
-        };
-        chat.borrow_mut().update_messages();
+        if let Some(_) = action.downcast_ref::<DownloadFileAction>() {
+            self.update_downloads();
```
noxware, line 276: Due to how de-normalized and duplicated is the downloads data, we still need to react and do this here to sync.

src/chat/chat_history.rs
joulei: Also seems unnecessary in my tests, I'd say safe to remove!

src/chat/chat_panel.rs
```diff
@@ -514,6 +496,28 @@ impl WidgetMatchEvent for ChatPanel {
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
         let store = scope.data.get_mut::<Store>().unwrap();
 
+        for action in actions {
+            if let Some(action) = action.downcast_ref::<ChatEntityAction>() {
+                if get_chat_id(store) == Some(action.chat_id) {
```
joulei, line 501: good catch!

src/data/chats/model_loader.rs
joulei: A very NIT about this (feel free to ignore) is that when I was reviewing `settings_screen.rs` went back to look up like: "Ok this is a ModelLoaderAction but which ModelLoaderAction was it".
Not sure if it's better to:
- Keep it a "generic" struct as you did.
- Make the action specific, e.g. with past tense: `ModelLoadedAction`
- ^ brings me to another option we use quite often, `ModelLoaderAction` as enum, with `None`, and `ModelLoaded` as variants. 
I don't love either 😅 just sharing my thoughts.

joulei (approved):
This looks great! It's a lot cleaner, more explicit and potentially a lot more efficient with all the redrawing we were doing before.
Thanks for leaving comments on the changes :)

Just left one very NIT comment, feel free to skip it.

src/data/chats/model_loader.rs
noxware: I admit model loader didn't change much , but you can see that in a positive way: It could mean that is very well encapsulated and didn't depend much on how we use actions at the current date. 

- In theory, an action from model loader would mean something has changed with it.
- In practice, that change it's reflected in it's status.
  - Status is already accessible thru model loader itself, so, an action doesn't need to carry any information apart from just notifying that a state transition occurred.

If it were to carry information, would be the status itself, or an enum as you mentioned representing the completed operation/transition over the state. I just pushed a commit doing the first one as a proof of concept :) Check https://github.com/moxin-org/moly/pull/295/commits/5f6e614dc246b9cb22a457ff9b7b7693af464666. As you can see it only "benefits" `settings_screen.rs`, but everything else doesn't. Give me your opinion about that. At least I would preserve the new name and the automatic post action on `set_status` as I now think it makes sense. But cloning the data into the action... I'm not sure if it's actually worth... See it for yourself and give me your honest opinion :)

PR #306, opened 2024-11-09
Allow stop button to work before any kind of response comes from the server

noxware:
# Context

Issue: https://github.com/moxin-org/moly/issues/206

# Changes

Take cancellation into account when:
- We still didn't receive the first part of the HTTP response.
- We still didn't drain and parse the full body stream when one-shot JSON response is expected. 

# Before

https://github.com/user-attachments/assets/8fec218c-ede2-468b-ac88-4c1f900cc3fc

# After

https://github.com/user-attachments/assets/d7d1bc7e-f501-4b74-a481-575158105c01


joulei (approved):
Nice use of tokio::select! :)

PR #312, opened 2024-11-18
make download_items responsive

Guocork:
Now, the layout in `download_items` is responsive.
截图 2024-11-18 16-14-41

### DIFF
--- a/src/landing/download_item.rs
+++ b/src/landing/download_item.rs
@@ -72,7 +72,7 @@ live_design! {
     }
 
     Progress = <View> {
-        width: 600,
+        width: Fill,
         height: Fit,
         spacing: 8,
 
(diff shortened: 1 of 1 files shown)


src/landing/download_item.rs
```diff
@@ -197,9 +197,23 @@ live_design! {
...
+            width: Fill,
+            height: Fit,
+            <Progress> {}
+        }
+
+        <View> {
+            width: Fill,
+            height: Fit,
+            <Actions> {}
+        }
```
joulei, line 216: While this works we can clean up a bit:
It seems that you can actually achieve this by simply changing the `width` and `height` property on those elements directly, instead of wrapping them in views.

You can keep:
```rs
<Information> {}
<Progress> {}
<Actions> {}
```

And update those values directly on each element:
```rs
Actions = <View> {
 width: Fill, height: Fit
// ...
}
```

Guocork: OK, I got it,thanks your reviewing.

joulei (approved):

src/landing/download_item.rs
joulei: Thanks!

PR #313, opened 2024-11-20
Main update 20 nov

joulei:


noxware (approved):
in mac os, `cargo run +nightly --release` works fine, no major problems detected from my quick testing.

I would merge this now and make a nightly release to test it in other devices. 

May be useful to do a full release later since we don't know if we will be able to release dev again quickly while stabilizing mofa branch there.

PR #320, opened 2024-11-29
Support @ model from prompt input

### LINKED ISSUE #308: Add support for @model for LLM models
## Context

For the agent integration with MoFa, we've added a `@name` selector capability within the chat input prompt that allows the user to select different agents. See https://github.com/moxin-org/moly/pull/279.

Image
Image

We now want to include this behaviour for LLM models within a chat.

## Requirements

- Writing `@` should bring up a list of models to select from (it should match the items and status of the model selector at the top of the chat panel.
- The user should be able to navigate the list via keyboard and select a model.
- Once the model is selected the UI should show the selection above the message input (see screenshot above).
- Sending a message with the newly selected model will load the model first and then send the message. Once the message is sent the text input should be reset to its initial state.
- The user should not be able to select the already current model, that item in the list should be shown as already selected.

noxware:
# Changes
- Support `@model` in prompt input.
- Support models in ~~AgentButton~~ EntityButton.
- Adjust when a model is remembered as the associated entity of a chat.
- Adjust when the model selector on the top shows its model as loading.
- Small compatibility changes between agents and models that impacted many files in the code base.

# Screenshots

<p align="center">
  
  
  
</p>

# Notes for reviewers
I tried to comment important changes, so it should be easier to notice important stuff and ignore collateral changes.

# Meta
Closes #308


moly-mofa/src/lib.rs
```diff
@@ -59,23 +59,21 @@ pub enum MofaAgent {
 }
 
 impl MofaAgent {
-    pub fn name(&self) -> String {
```
noxware, line 62: Premature usage of `String` type here forces the use of some `.clone()`s around the code base that could be easily avoided.

src/data/chats/chat_entity.rs
```diff
@@ -0,0 +1,38 @@
+//! Hopefully, provisional solution to unify model files and agents in the chat system.
+
+use moly_mofa::MofaAgent;
+use moly_protocol::data::{File, FileID};
+use serde::{Deserialize, Serialize};
+
+/// Identifies either a model file or an agent.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum ChatEntityId {
```
noxware, line 9: This type is actually what was called `ChatEntity` in `chat.rs`. Renamed it to `Id` as it's not intended to own the entity itself. Agents are currently enums, so they are identified by themselves, not like files.

src/data/chats/chat_entity.rs
```diff
@@ -0,0 +1,38 @@
...
+    ModelFile(FileID),
+    /// Since agents are currently fixed enum values, the agent itself is the identifier.
+    Agent(MofaAgent),
+}
+
+/// Reference to either a model file or an agent.
+///
+/// Can be used to chain iterators of both types or simply to take either as a parameter.
+#[derive(Debug, Clone, Serialize, Copy)]
+pub enum ChatEntityRef<'a> {
```
noxware, line 19: After some cleanup this is only used at `prompt_input.rs` and `entity_button.rs`.

src/data/chats/chat_entity.rs
```diff
@@ -0,0 +1,38 @@
+//! Hopefully, provisional solution to unify model files and agents in the chat system.
```
noxware, line 1: Just moved these glues to a separate file hoping we find a way to really unify models and agents later.

src/shared/actions.rs
```diff
@@ -1,16 +1,11 @@
 use makepad_widgets::{ActionDefaultRef, DefaultNone};
-use moly_mofa::MofaAgent;
 use moly_protocol::data::FileID;
 
-#[derive(Clone, Debug)]
-pub enum ChatHandler {
```
noxware, line 6: This type is equivalent to `ChatEntityId` (ex-`ChatEntity`), so I removed it and replaced its usage.

src/data/chats/chat_entity.rs
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+
+    pub fn name(&self) -> &str {
+        match self {
+            ChatEntityRef::ModelFile(file) => &file.name,
+            ChatEntityRef::Agent(agent) => agent.name(),
+        }
+    }
+}
```
noxware, line 38: Note: The presence of `ChatEntityId` and `ChatEntityRef` implies the existence of a `ChatEntity` type, thas was not needed for this work, but it would be something that really owns the model or agent, not just an id or ref.

src/chat/entity_button.rs
```diff
@@ -1,9 +1,10 @@
+use crate::data::chats::chat_entity::{ChatEntityId, ChatEntityRef};
```
noxware, line 1: This was `agent_button.rs`. Modified to be able to handle both, agents and models.

Also, removed the flags used to change the internal behavior of the widget and removed the globally dispatched actions caused by those flags. Instead, a simply `clicked()` method has been added, and handling occurs on the parent maximizing reusability and simplifying the internals.

src/chat/entity_button.rs
```diff
@@ -100,71 +101,105 @@ live_design!(
...
     view: View,
 
     #[rust]
-    agent: Option<MofaAgent>,
-
-    #[live(false)]
-    create_new_chat: bool,
-
-    #[live(false)]
-    select_agent_on_prompt: bool,
```
noxware, line 114: This boolean flags change the widget behavior to do stuff that can be easily done outside. Those behaviors are implemented internally by dispatching global actions which harms re-usability unnecessarily. 

So they were replaced with a simple `.clicked()` method (like normal button's method), and that stuff is handled outside.

src/chat/entity_button.rs
```diff
@@ -100,71 +101,105 @@ live_design!(
...
     }
-}
 
-impl AgentButton {
-    pub fn set_agent(&mut self, agent: &MofaAgent, show_description: bool) {
+    pub fn get_entity_id(&self) -> Option<&ChatEntityId> {
+        self.entity.as_ref()
+    }
+
+    pub fn set_agent(&mut self, agent: &MofaAgent) {
```
noxware, line 137: Since most usage of this component is still to display agents, this method is preserved but implemented on top of `set_entity`. This avoid propagating `ChatEntityRef` unnecessarily across the code base.

src/chat/model_selector.rs
```diff
@@ -1,5 +1,5 @@
 use crate::{
```
noxware, line 1: Changes in this file are about not showing the model in the selector as "loading" if it's not the one actually being loaded.

For example, if the chat has an associated model A, and you used `@` to invoke another model temporally.

src/chat/prompt_input.rs
```diff
@@ -1,19 +1,19 @@
 use makepad_widgets::*;
```
noxware, line 1: "Large diffs are not rendered by default. " he... Of course, this file was the most involved with the issue.

But most work to support models here has been done in `entity_button.rs` (ex-`agent_button.rs`) and by `ChatEntityRef`. And some stuff has been simplified. For example, using `items()` from the list widget allows you directly know which item has been selected without reconstructing the whole list again.

So, actually, many changes here are just deleting the world `agent` with quick searches from many places like method names and so on.

src/data/chats/chat.rs
```diff
@@ -31,18 +32,12 @@ enum ChatEntityActionKind {
     MofaAgentCancelled,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub enum ChatEntity {
```
noxware, line 35: This is now `ChatEntityId`.

src/data/chats/mod.rs
```diff
@@ -67,15 +69,6 @@ impl Chats {
     pub fn load_model(&mut self, file: &File, override_port: Option<u16>) {
         self.cancel_chat_streaming();
 
-        if let Some(mut chat) = self.get_current_chat().map(|c| c.borrow_mut()) {
```
noxware, line 70: Not every time we load a model we mean to save it as the associated one. When using `@` to invoke a model, it will not be saved as an associated one. 
So this decision must happen outside of this method.

src/data/store.rs
```diff
@@ -149,40 +150,41 @@ impl Store {
...
-            _ => {
-                self.send_chat_message(prompt, regenerate_from);
-            }
+        if let Some(entity_id) = entity_id {
+            self.send_entity_message(&entity_id, prompt, regenerate_from);
         }
     }
 
-    pub fn send_chat_message(&mut self, prompt: String, regenerate_from: Option<usize>) {
+    pub fn send_entity_message(
```
noxware, line 163: Some refactoring here. This method handles what `send_chat_message` and `send_agent_message` handled, but allowing to specify a specific model, and also covering the last part of the `send_message_to_current_entity` method.

src/data/chats/chat.rs
```diff
@@ -331,12 +326,11 @@ impl Chat {
             id: next_id + 1,
             role: Role::Assistant,
             username: Some(wanted_file.name.clone()),
-            entity: Some(ChatEntity::ModelFile(wanted_file.id.clone())),
+            entity: Some(ChatEntityId::ModelFile(wanted_file.id.clone())),
             content: "".to_string(),
         });
 
         self.state = ChatState::Receiving;
-        self.associated_entity = Some(ChatEntity::ModelFile(wanted_file.id.clone()));
```
noxware, line 339: This line may pass unnoticed so I'm commenting it. Now, sending a message will not override the associated entity of the chat as `@` invoked models should not override the currently selected one on the selector.

joulei (commented):
Great job! 👏🏽 
Leaving some comments about the UI texts.

Also, I think we should section the list between agents and models. 
Currently agents have a "profile picture" while the models don't, it makes the UI feel a bit unfinished (or look like some things haven't loaded). Perhaps visually separating the agents and models makes it look a lot better.
We can do something like label the sections: (e.g. the added "Models" text here, we would the same for agents)

image
We could also add a horizontal line (like in the selector) if needed, but it might be cleaner if we just title the sections. What do you think?

joulei (approved):
Heads-up, I pushed a commit making those minor visual changes I mentioned.
I'll merge so that I can prepare a demo on top of this, thanks!

PR #322, opened 2024-12-04
Fix chat switching

noxware:
# Initial problem

I noticed that when switching from a long chat to a new agent chat, portal list bugs due to scroll reset not being triggered. Some dev commits ago this caused the app to crash but now it just causes a visual and performance issue.

# Deeper problem

Actually, this issue can be provoked by mixing other kinds of user interaction:
- If you are in a long chat, then you go to "My Models" and then you start a new chat with the same model from there.
- If you have 2 chats, one long and one short, and being at the long one you delete it.

And there may be more I'm not noticing now. Or at least, there will be more. 

# Solution

- Do not try to react to every possible action that my update current chat id.
- Instead, directly react to the chat id change itself :)

There are two ways to do this without a reactivity abstraction:
- Track chat id in a widget local field and compare it to determine if it changed.
  - This is the same approach that was used for `chat_params.rs`.
  - And is the approach I'm currently proposing in this PR due to it's simplicity and being proven to work somewhere else. 
- Alternatively, dispatch a specific global action from `set_current_chat` itself like `CurrentChatIdChanged` using `Cx::post_action`.

No matter the approach,  `ChatHistoryCardAction::ChatSelected` becomes unnecessary and can be safely deleted.

### DIFF
--- a/src/chat/chat_panel.rs
+++ b/src/chat/chat_panel.rs
@@ -17,8 +17,8 @@ use crate::{
 };
 
 use super::{
-    chat_history_card::ChatHistoryCardAction, model_selector_list::ModelSelectorListAction,
-    prompt_input::PromptInputWidgetExt, shared::ChatAgentAvatarWidgetRefExt,
+    model_selector_list::ModelSelectorListAction, prompt_input::PromptInputWidgetExt,
+    shared::ChatAgentAvatarWidgetRefExt,
 };
 
 live_design! {
--- a/src/chat/chat_panel.rs
+++ b/src/chat/chat_panel.rs
@@ -394,15 +394,25 @@ pub struct ChatPanel {
 
     #[rust(false)]
     focus_on_prompt_input_pending: bool,
+
+    #[rust]
+    current_chat_id: Option<ChatID>,
 }
 
 impl Widget for ChatPanel {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        let store = scope.data.get_mut::<Store>().unwrap();
+
+        if store.chats.get_current_chat_id() != self.current_chat_id {
+            self.current_chat_id = store.chats.get_current_chat_id();
+            self.reset_scroll_messages(store);
+            self.redraw(cx);
+        }
+
         self.view.handle_event(cx, event, scope);
         self.widget_match_event(cx, event, scope);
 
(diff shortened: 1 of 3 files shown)


src/chat/chat_panel.rs
```diff
@@ -762,7 +766,7 @@ impl ChatPanel {
         }
 
         // Let's confirm we're in an appropriate state to send a message
-        self.update_state(store);
+        self.update_state(scope);
```
noxware, line 769: This call and the if should be technically unnecessary. Like, how did you trigger a `send_message` if you are at a state where the UI should be disabled? If we can, there is a bug.

But yeah, let's just adapt the code as necessary and preserve it just in case.

If we do work about making chats embedded, this may be re-implemented anyways.

src/chat/chat_panel.rs
```diff
@@ -394,15 +394,25 @@ pub struct ChatPanel {
 
     #[rust(false)]
     focus_on_prompt_input_pending: bool,
+
+    #[rust]
+    current_chat_id: Option<ChatID>,
```
noxware, line 399: Used to check if the current chat changed.

src/chat/chat_panel.rs
```diff
@@ -394,15 +394,25 @@ pub struct ChatPanel {
...
+            self.reset_scroll_messages(store);
+            self.redraw(cx);
+        }
+
         self.view.handle_event(cx, event, scope);
         self.widget_match_event(cx, event, scope);
 
-        let store = scope.data.get_mut::<Store>().unwrap();
-        self.update_state(store);
+        self.update_state(scope);
```
noxware, line 415: Mainly changed to pass scope and not store down, or it would need to get the store 2 times. You can't just use one cause the store would be needed at the start and at the end, but the middle instructions requires scope, which would cause borrowing issues.

joulei (approved):
Thanks for the breakdown!
I think in the future we might want to think the actions the other way around (opposite to what we had before these changes). Instead of a `SomewhereElse::SomethingHappened` action that other parts of the app should check to update their state, the action should be `ThingThatShouldUpdate::PleaseUpdateX`, or as you mention, update it in the store and then notify the rest of the app `ChatIdChanged` - `ActiveChatChanged`, etc.

I think checking if the id changed within `ChatPanel` should be enough for now. We're using a similar pattern in most of Robrix, and while a bit tedious, it shouldn't be that much less efficient than using actions.

PR #323, opened 2024-12-06
Fix chat switch streaming cancel

noxware:
# Problem

When switching to a different chat from one that was streaming, it will not always cancel the previous streaming causing issues in the UI.

Some examples of cancellation not happening are when creating a new chat or when deleting the streaming chat.

### DIFF
--- a/src/data/chats/mod.rs
+++ b/src/data/chats/mod.rs
@@ -210,7 +210,6 @@ impl Chats {
 
     pub fn remove_chat(&mut self, chat_id: ChatID) {
         if self.current_chat_id == Some(chat_id) {
-            self.cancel_chat_streaming();
             self.set_current_chat(self.get_last_selected_chat_id());
         }
 
--- a/src/data/chats/mod.rs
+++ b/src/data/chats/mod.rs
@@ -209,16 +209,19 @@ impl Chats {
     }
 
     pub fn remove_chat(&mut self, chat_id: ChatID) {
-        if let Some(chat) = self.saved_chats.iter().find(|c| c.borrow().id == chat_id) {
-            chat.borrow().remove_saved_file();
-        };
-        self.saved_chats.retain(|c| c.borrow().id != chat_id);
-
-        if let Some(current_chat_id) = self.current_chat_id {
-            if current_chat_id == chat_id {
-                self.set_current_chat(self.get_last_selected_chat_id());
-            }
+        if self.current_chat_id == Some(chat_id) {
+            self.cancel_chat_streaming();
+            self.set_current_chat(self.get_last_selected_chat_id());
         }
+
+        let pos = self
+            .saved_chats
+            .iter()
+            .position(|c| c.borrow().id == chat_id)
+            .expect("non-existing chat");
+
+        let chat = self.saved_chats.remove(pos);
+        chat.borrow().remove_saved_file();
     }
 
     pub fn handle_action(&mut self, action: &Box<dyn ActionTrait>) {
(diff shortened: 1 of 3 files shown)


src/data/chats/mod.rs
```diff
@@ -99,13 +100,15 @@ impl Chats {
         self.saved_chats.iter().find(|c| c.borrow().id == chat_id)
     }
 
-    pub fn set_current_chat(&mut self, chat_id: ChatID) {
+    pub fn set_current_chat(&mut self, chat_id: Option<ChatID>) {
```
noxware, line 103: As long as chat switching happens thru this function, the problem should not happen.

src/data/chats/mod.rs
```diff
@@ -194,26 +198,29 @@ impl Chats {
 
     pub fn create_empty_chat_and_load_file(&mut self, file: &File) {
         let mut new_chat = Chat::new(self.chats_dir.clone());
+        let id = new_chat.id;
         new_chat.associated_entity = Some(ChatEntityId::ModelFile(file.id.clone()));
         new_chat.save();
 
-        self.current_chat_id = Some(new_chat.id);
         self.saved_chats.push(RefCell::new(new_chat));
+        self.set_current_chat(Some(id));
```
noxware, line 206: The chat must exist in saved messaged before setting it.

src/data/chats/mod.rs
```diff
@@ -194,26 +198,29 @@ impl Chats {
...
-            chat.borrow().remove_saved_file();
-        };
-        self.saved_chats.retain(|c| c.borrow().id != chat_id);
-
-        if let Some(current_chat_id) = self.current_chat_id {
-            if current_chat_id == chat_id {
-                self.current_chat_id = self.get_last_selected_chat_id();
-            }
+        if self.current_chat_id == Some(chat_id) {
+            self.set_current_chat(self.get_last_selected_chat_id());
```
noxware, line 213: This is the reason why I made set_current_chat argument an Option.

src/data/chats/mod.rs
```diff
@@ -194,26 +198,29 @@ impl Chats {
 
     pub fn create_empty_chat_and_load_file(&mut self, file: &File) {
         let mut new_chat = Chat::new(self.chats_dir.clone());
+        let id = new_chat.id;
         new_chat.associated_entity = Some(ChatEntityId::ModelFile(file.id.clone()));
         new_chat.save();
```
noxware, line 203: I'm concerned about file writes happening more than necessary here as writes to disk are a precious resource.

I can remove this and other `save` statements as they are performed by `set_current_chat` internally. But would not be clear here that the chat is being saved (without a comment), and there is the risk of removing the `save` on `set_current_chat` affecting this other places).

Another approach would be to buffer writes, so if you do many `.save()` in a second, only the last one is flushed. I tried adding a function that does exactly this (see `buffered_write_to_file` in `filesystem.rs`) in the simplest and smallest way I could think about. But doing that properly would require async code, separate tasks per file, non-blocking error handling, etc.

Please let me know if you don't agree on adding that function and I will remove it, no question asked.

src/data/filesystem.rs
```diff
@@ -80,3 +85,37 @@ pub fn write_to_file(path: PathBuf, json: &str) -> Result<(), std::io::Error> {
     file.write_all(json.as_bytes())?;
     Ok(())
 }
+
+static BUFFERED_WRITER: LazyLock<Sender<(PathBuf, String)>> = LazyLock::new(|| {
```
noxware, line 89: `LazyLock` is supported in Rust stable since `1.80`. Seems like the build CI is using an older Rust version?

joulei (approved):

PR #324, opened 2024-12-06
Buffered save

noxware:


src/data/filesystem.rs
```diff
@@ -80,3 +85,37 @@ pub fn write_to_file(path: PathBuf, json: &str) -> Result<(), std::io::Error> {
     file.write_all(json.as_bytes())?;
     Ok(())
 }
+
+static BUFFERED_WRITER: LazyLock<Sender<(PathBuf, String)>> = LazyLock::new(|| {
```
noxware, line 89: LazyLock is supported in Rust stable since 1.80. Seems like the build CI is using an older Rust version?

PR #325, opened 2024-12-11
Support multiple MoFa agents and servers

joulei:
## Changes
This PR adds support for multiple MoFa severs and multiple agents.
I've extended the settings UI to support adding servers at runtime, and testing the connection.

## Limitations
- Currently we don't support editing a server address, this is because modifying a server address while in use adds a lot of complexity (some kind of mutex around the address, agent list diff reconciliation, agent id reconciliation, etc.). We could explore this in the future.
- There are some minor improvements we can tackle separately: 
    - Add confirmation dialog when deleting a server
    - Supporting server shutdown and turn on - we could keep the servers on the list but turn them on/off on demand
    - Periodically check for server connection health and agent list refresh
- Most MoFa server don't yet support sending an actual list of agents on the `v1/models` endpoint, therefore we're still hardcoding a Reasoner agent.

## Screen recordings

Adding a new server and talking to an agent:

https://github.com/user-attachments/assets/c8f1657b-90f3-4ad9-ac3a-e59b6b610d10

Handling of no agents, disconnected servers or no servers:

https://github.com/user-attachments/assets/4ec5b1ff-2da9-48a0-bae7-8769ab94c101


src/data/chats/mod.rs
```diff
@@ -225,4 +261,113 @@ impl Chats {
...
+            }
+            Ok(MofaServerResponse::Unavailable(server_address)) => {
+                Cx::post_action(MoFaTestServerAction::Failure(Some(server_address)));
+            }
+            Err(e) => {
+                error!("Error receiving response from MoFa backend: {:?}", e);
+                Cx::post_action(MoFaTestServerAction::Failure(None));
+            }
+        });
+    }
```
joulei, line 331: Combined server testing and agent fetching into a single operation to simplify async flow.

Previously, we had separate operations for testing server connectivity and fetching agents (which would both just do the same thing, fetching `/v1/models`). 
This created complexity in managing multiple async operations without proper abstractions for 
state management. By combining these into a single operation, we:

1. Reduce complexity in handling async state transitions
2. Maintain consistency between server status and available agents
3. Simplify the UI feedback loop
4. Make the transition to real server responses easier when they become available

src/data/chats/chat.rs
```diff
@@ -398,15 +397,12 @@ impl Chat {
...
         prompt: String,
-        mae_backend: &MofaBackend,
+        mofa_client: &MofaClient,
     ) {
         let (tx, rx) = mpsc::channel();
-        mae_backend
-            .command_sender
-            .send(SendTask(prompt.clone(), agent.clone(), tx.clone()))
-            .expect("failed to send message to agent");
+        mofa_client.send_message_to_agent(&agent, &prompt, tx);
```
joulei, line 406: You may notice that in contrast to `send_message_to_model` and other similar operations, I'm not talking to a command sender directly but rather simply calling a method in the client.
I think having an indirection like this is clearer. The channel command handling is actually just a limitation of our thread and async handling (related to the client), not a meaningful part of the models or agents interface. I think previously we "leaked" those specifics into different parts of the UI and they just make the code harder to read and follow.
This will be slightly inconsistent for now but I think it's worth it, and either way we're looking into refactoring many aspects of this.

src/data/chats/mod.rs
```diff
@@ -5,43 +5,75 @@ pub mod model_loader;
...
     pub model_loader: ModelLoader,
 
+    pub mofa_servers: HashMap<MofaServerId, MofaServer>,
+    pub available_agents: HashMap<AgentId, MofaAgent>,
+
     current_chat_id: Option<ChatID>,
     chats_dir: PathBuf,
 
     override_port: Option<u16>,
 }
```
joulei, line 63: I don't love that I had to put these here. (also in generally I don't love polluted `mod.rs` files)
Originally it felt more appropriate to put these `HashMaps` in Store and have the related operations there (or in a separate module). But due to how the `ChatPanel` and other parts of the UI interact with `/chats/mod.rs` I had to keep everything together for the sake of sanity and have some consistency.
I think after this PR we can think of some refactoring, I don't think this `Chats` abstraction is the best, and there's some mix up between this and similar things in the `Store`.
Somewhat related we can also add some traits for unifying the MoFa and Models operations. I didn't want to make this PR harder to review.

src/chat/chat_panel.rs
```diff
@@ -506,8 +506,14 @@ impl WidgetMatchEvent for ChatPanel {
...
                     }
-                    ChatEntityId::Agent(agent) => {
-                        store.chats.create_empty_chat_with_agent(agent);
+                    ChatEntityId::Agent(agent_id) => {
+                        let agent = store
+                            .chats
+                            .available_agents
+                            .get(&agent_id)
+                            .cloned()
+                            .unwrap_or_default();
```
noxware, line 515: mmm i think it may be better to ensure we can't start a chat with an unrecognized agent, and panic or eprintln something here to indicate a dev mistake

src/chat/chat_history_card.rs
```diff
@@ -360,8 +360,14 @@ impl Widget for ChatHistoryCard {
...
 
-        match chat.borrow().associated_entity {
-            Some(ChatEntityId::Agent(agent)) => {
+        match &chat.borrow().associated_entity {
+            Some(ChatEntityId::Agent(agent_id)) => {
+                let agent = store
+                    .chats.available_agents
+                    .get(&agent_id)
+                    .cloned()
+                    .unwrap_or_default();
```
noxware, line 369: I would divide this flow to handle the case of agent missing, or at least, if you really want to do this, I would make a constructor `MofaAgent::unknown()` or something similar instead of using the default as that

choose your pill

src/chat/entity_button.rs
```diff
@@ -135,7 +154,7 @@ impl EntityButton {
     }
 
     pub fn set_agent(&mut self, agent: &MofaAgent) {
-        self.set_entity(ChatEntityRef::Agent(agent));
+        self.set_entity(ChatEntityRef::Agent(&agent));
```
noxware, line 157: `agent` is already of type `&MofaAgent` so the added `&` is not necessary

src/chat/model_selector_item.rs
```diff
@@ -126,9 +126,9 @@ impl ModelSelectorItemRef {
         inner.entity = ModelSelectorEntity::Model(model);
     }
 
-    pub fn set_agent(&mut self, agent: MofaAgent) {
+    pub fn set_agent(&mut self, agent: &MofaAgent) {
         let Some(mut inner) = self.borrow_mut() else { return };
-        inner.entity = ModelSelectorEntity::Agent(agent);
+        inner.entity = ModelSelectorEntity::Agent(agent.clone());
```
noxware, line 131: You want to work with an owned `MofaAgent`, so keep taking an owned `MofaAgent` :P

I would avoid cloning inside the function if possible and let the caller clone if it's really necessary.

Does inverting the two last lines work?

```rust
pub fn set_agent(&mut self, agent: MofaAgent) {
    let Some(mut inner) = self.borrow_mut() else {
        return;
    };
    inner.chat_agent_avatar(id!(avatar)).set_agent(&agent);
    inner.entity = ModelSelectorEntity::Agent(agent);
}
```

noxware (commented):
I have to pause for a moment, I will finish the review when back

src/data/chats/mod.rs
```diff
@@ -225,4 +261,113 @@ impl Chats {
...
+
+    /// Tests the connection to a MoFa server by requesting /v1/models
+    /// The connection status is updated at the App level based on the actions dispatched
+    /// 
+    /// We perform fetching (and therefore testing) as follows:
+    /// 1. We create a channel used to communicate with a background thread that will do the http request
+    /// 2. We start that background thread and instruct it to fetch the agents from the server
+    /// 3. The thread sends a message back to the channel created here with the server response
+    /// 4. We forward that result back to the UI thread using Cx::post_action
+    /// 5. The action is handled at the App level, which tells the store to update the connection status and agent list
```
noxware, line 311: I would be careful with long doc comments. Unless you are making a library, long comments are likely to become unmaintained. 

I think if one is curious of how this works inside, that person will look inside. So documenting step by step how it works inside adds little value. 

I would preserve the "what it does", and "how is the result reflected":

```rust
/// Tests the connection to a MoFa server by requesting /v1/models.
///
/// The connection status is updated at the App level based on the actions dispatched.
```

src/data/store.rs
```diff
@@ -1,25 +1,23 @@
 use super::chats::chat::ChatID;
 use super::chats::chat_entity::ChatEntityId;
 use super::chats::model_loader::ModelLoaderStatusChanged;
+use super::chats::MoFaTestServerAction;
```
noxware, line 4: If `MoFa`, then `mo_fa`?

src/data/store.rs
```diff
@@ -172,8 +160,15 @@ impl Store {
...
-                ChatEntityId::Agent(agent) => {
-                    chat.send_message_to_agent(*agent, prompt, &self.mofa_backend);
+                ChatEntityId::Agent(agent_id) => {
+                    if let (Some(client), Some(agent)) = (
+                        self.chats.get_client_for_agent(agent_id),
+                        self.chats.available_agents.get(agent_id)
+                    ) {
+                        chat.send_message_to_agent(agent, prompt, &client);
+                    } else {
+                        println!("client or agent not found: {:?}", agent_id);
```
noxware, line 170: nitpick: I would print warnings, errors and debug messages to `stderr` with `eprintln!`.

src/landing/model_list.rs
```diff
@@ -149,14 +158,15 @@ impl Widget for ModelList {
     }
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
-        let store = scope.data.get::<Store>().unwrap();
-        let agents = MofaBackend::available_agents();
+        let store = scope.data.get_mut::<Store>().unwrap();
```
noxware, line 161: `get_mut` but variable is not mutable, so probably unnecessary

src/landing/model_list.rs
```diff
@@ -166,12 +176,20 @@ impl Widget for ModelList {
...
                 items.push(Item::Header("Featured Agents"));
-                items.extend(agents.chunks(3).map(|chunk| Item::AgentRow {
-                    agents: chunk,
-                    margin_bottom: 15.0,
-                }));
-                if let Some(Item::AgentRow { margin_bottom, .. }) = items.last_mut() {
-                    *margin_bottom = 0.0;
+                let agents_availability = store.chats.are_agents_available();
+                match agents_availability {
+                    AgentsAvailability::NoServers => items.push(Item::NoAgentsWarning("Not connected to any MoFa servers.")),
```
noxware, line 181: hmmm... i suppose these strings are the same as in `chat_history`, maybe you could add a `to_human_message()`-like function to `AgentsAvailability` if you want to dedup, but I also like messages defined in the UI itself so, don't mind, just thinking aloud

src/settings/mofa_settings.rs
```diff
@@ -11,210 +13,347 @@ live_design! {
...
+
+    ICON_SUCCESS = dep("crate://self/resources/images/circle_check_icon.png")
+    ICON_LOADER = dep("crate://self/resources/images/loader_icon.png")
+    ICON_FAILURE = dep("crate://self/resources/images/refresh_error_icon.png")
+
+    // Not making this based on <Icon> because button does not support images
+    // (and these SVGs are too complex for Makepad's SVG support)
+    ConnectionStatusIcon = <View> {
+        visible: false
+        cursor: Hand
```
noxware, line 28: So this as a whole is a button, not an icon, right?

src/settings/mofa_settings.rs
```diff
@@ -11,210 +13,347 @@ live_design! {
...
         flow: Down
-        spacing: 20
 
-        <Label> {
-            draw_text:{
-                text_style: <BOLD_FONT>{font_size: 16}
-                color: #000
+        separator = <View> {
+            margin: {left: 20, right: 20, top: 0, bottom: 10}
+            width: Fill,
```
noxware, line 59: stupid comment crossing my mind: views are Fill, Fill by default. We are not being explicit about height being Fill, but we are explicit on width being Fill. Shall we remove `width: Fill`? add height?

src/settings/mofa_settings.rs
```diff
@@ -11,210 +13,347 @@ live_design! {
...
+                        let template = live_id!(mofa_server_item);
+                        let item = list.item(cx, item_id, template);
+
+                        // hide the separator for the first item
+                        if item_id == 0 {
+                            item.view(id!(separator)).set_visible(false);
+                        }
+
+                        let server = &servers[item_id].clone();
+                        let mut item_scope = Scope::with_props(server);
```
noxware, line 269: :o

noxware (commented):
ufff took me a lot longer than I expected but it's done, not the perfect review but it's a review :D

moly-mofa/src/lib.rs
```diff
@@ -273,10 +274,12 @@ pub enum ChatResponse {
...
 
-// ====
-
 #[derive(Clone, Debug, Serialize, Deserialize)]
 struct ChatRequest {
     model: String,
     messages: Vec<MessageData>,
 }
+
+fn unique_agent_id(agent_id: &AgentId, server_address: &str) -> AgentId {
```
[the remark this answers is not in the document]
joulei: > double thinking, this can even be the `new` constructor of `AgentId`
> 
> ```rust
> AgentId::new("whatever", "http://hmm.com");
> ```

Good catch! 🙌🏽

src/settings/mofa_settings.rs
joulei: I guess we can come up with a convention about that (and other DSL things like comma usage, etc.
In this case the height is 1 so both are explicit, I'll remove the width for now.

src/chat/chat_panel.rs
joulei: Answering on the next one

src/chat/chat_history_card.rs
joulei: So this case is not about starting new chats with an inaccesible agent but rather about how we handle the chat history whenever we're referring to an agent that is no longer reachable.

I agree that this was not very clean or clear, I've updated (across all the files) to:
```rs
let agent = store.chats.get_agent_or_placeholder(&agent_id);
```
With:
```rs
/// Returns a reference to an agent by ID, falling back to an unknown agent placeholder
/// if the agent is not found in the available agents list.
/// 
/// This is useful when dealing with historical chat references to agents that may
/// no longer be available (e.g., server disconnected or agent deleted).
/// 
/// In the future, we'll want a more sophisticated solution, by potentially storing 
/// agent information locally and updating it when a server is connected.
pub fn get_agent_or_placeholder(&self, agent_id: &AgentId) -> &MofaAgent {
    self.available_agents.get(agent_id).unwrap_or(&self.unknown_agent)
}
```

As mentioned there this is a temporary solution while we wait for confirmation from the MoFa team on what information they will provide, how could we keep a local copy of some data, sync, etc. Right now you cannot start a new chat with an unreachable agent which is the most important part.
Besides that, we should improve the error handling (mostly notify the user properly) of errors in either WasmEdge backend models or MoFa. Currently for either of them you just don't see anything happening on the screen, but we should fix that for both cases, perhaps in a separate PR.

src/data/chats/mod.rs
```diff
@@ -5,43 +5,75 @@ pub mod model_loader;
...
-use std::sync::mpsc::channel;
+use std::sync::mpsc::{self, channel};
 use std::{cell::RefCell, path::PathBuf, rc::Rc};
 
 use super::filesystem::setup_chats_folder;
 
+#[derive(Clone, Debug)]
+pub struct MofaServer {
+    pub address: String,
+    pub client: Rc<MofaClient>,
```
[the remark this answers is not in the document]
joulei: Oops, leftover from me trying to emulate what's done for the WasmEdge backend, but this evolved in a different direction. Removed.

moly-mofa/src/lib.rs
```diff
@@ -50,117 +50,122 @@ pub struct MofaResponseSearchAssistant {
...
+    /// (due to the server not being available, the server no longer providing the agent, etc.).
+    fn default() -> Self {
+        MofaAgent {
+            id: AgentId("default".to_string()),
+            name: "Inaccesible Agent".to_string(),
+            description: "Default".to_string(),
+            agent_type: AgentType::Reasoner,
+            server_id: MofaServerId("default".to_string()),
+        }
+    }
```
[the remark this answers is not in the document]
joulei: Replied in another comment.

src/data/chats/mod.rs
```diff
@@ -5,43 +5,75 @@ pub mod model_loader;
...
 use std::fs;
-use std::sync::mpsc::channel;
+use std::sync::mpsc::{self, channel};
 use std::{cell::RefCell, path::PathBuf, rc::Rc};
 
 use super::filesystem::setup_chats_folder;
 
+#[derive(Clone, Debug)]
+pub struct MofaServer {
+    pub address: String,
```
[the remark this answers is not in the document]
joulei: Fixed

src/data/chats/mod.rs
joulei: Yeah I think at the time I was trying to make clear what the changes are and I really think some explaining is needed because this mechanism of fetching and updating store/UI is quite convoluted. 

> I think if one is curious of how this works inside, that person will look inside. So documenting step by step how it works inside adds little value.

I don't think this is true here because what happens and how it happens is scattered across the application and some doc could save a lot of time. In most platforms if you say `bla.request_something` you usually get a response, not a bunch of side effects. 
However I do think this fits better in some kind of higher level documentation rather than in a specific method. 
(I should have just put this as a comment on the PR, just wanted to make reviewing easier actually).

I've shortened it with your suggestion :)

moly-mofa/src/lib.rs
```diff
@@ -50,117 +50,122 @@ pub struct MofaResponseSearchAssistant {
...
         }
     }
 
-    pub fn main_loop(command_receiver: mpsc::Receiver<MofaAgentCommand>) {
-        println!("MoFa backend started");
+    /// Handles the communication between the MofaClient and the MoFa server.
+    /// 
+    /// This function runs in a separate thread and processes commands received through the command channel.
+    /// 
+    /// The loop continues until the command channel is closed or an unrecoverable error occurs.
```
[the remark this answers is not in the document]
joulei: Good catch! I actually didn't touch that (we're handling like this in a lot of places). But I wrote that doc with AI assistance and didn't catch :)
I'm switching to 
```rs
while let Some(message) = command_receiver.recv() { ... }
```

noxware (approved):
lgtm

PR #333, opened 2024-12-27
Some mofa fixes

noxware:
Changes

- Fix mofa client fake impl
- Always insert default mofa server
- Also, some auto-formatting

### DIFF
--- a/moly-mofa/src/lib.rs
+++ b/moly-mofa/src/lib.rs
@@ -86,7 +86,8 @@ impl MofaAgent {
         MofaAgent {
             id: AgentId("unknown".to_string()),
             name: "Inaccesible Agent".to_string(),
-            description: "This agent is not currently reachable, its information is not available".to_string(),
+            description: "This agent is not currently reachable, its information is not available"
+                .to_string(),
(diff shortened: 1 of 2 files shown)


moly-mofa/src/lib.rs
```diff
@@ -229,21 +239,41 @@ impl MofaClient {
...
+            Self::process_agent_commands(command_receiver, address_clone);
+        });
+
+        Self {
+            command_sender,
+            address,
+        }
+    }
+
+    fn new_fake(address: String) -> Self {
```
noxware, line 255: main changes here

src/data/store.rs
```diff
@@ -93,10 +92,10 @@ impl Store {
         store.init_current_chat();
 
         store.search.load_featured_models();
+        store
+            .chats
+            .register_mofa_server(DEFAULT_MOFA_ADDRESS.to_string());
 
-        if moly_mofa::should_be_real() && moly_mofa::should_be_visible() {
```
noxware, line 97: should always happen

joulei (approved):

Issue #199, opened 2024-08-06
Display spinning/loading animation while an LLM response message is still streaming


kevinaboos:
## Problem
Currently, the UI makes it impossible to tell if a model has finished streaming its response back to the user, or if it is still underway and is just taking a long time to calculate the response. This is especially prevalent on slower machines without AVX-512 or CUDA, and even more of a problem when the message being returned by the model is very very long.

## Proposed fix
To improve the user experience, a small spinning loading wheel (or some other kind of animation) should be added either beneath or next to the latest message in the chat view, which will spin or animate in some way that makes it obvious that the model is still generating a streaming a response. Then, once the message streaming has completed, the spinning animation would end and set itself to non-visible.

## Additional improvements
Alternatively, that spinning animation could change to a tiny checkmark icon or something similar, in order to indicate that the message had been fully streamed from the model.

In addition, this small loading animation could also be displayed while a model is being loaded in the background, which again can take quite a long time (20+ seconds) on slower machines. This would help inform the user that something is indeed still happening in the background, rather than leaving the user confused while the UI does nothing for dozens of seconds.

## Implementation starter ideas
The wasmedge frontend/backend does make it obvious when a message has completed being streamed. You can observe this while looking at the console log:
```
[2024-08-06T21:56:32Z INFO  llama-core] Get output buffer generated by the model named Meta-Llama-3-8B-Instruct-f16.gguf in the stream mode.
[2024-08-06T21:56:33Z INFO  llama-core] Get output buffer generated by the model named Meta-Llama-3-8B-Instruct-f16.gguf in the stream mode.
[2024-08-06T21:56:33Z INFO  llama-core] Get output buffer generated by the model named Meta-Llama-3-8B-Instruct-f16.gguf in the stream mode.
[2024-08-06T21:56:33Z INFO  llama-core] Get output buffer generated by the model named Meta-Llama-3-8B-Instruct-f16.gguf in the stream mode.
[2024-08-06T21:56:34Z INFO  llama-core] Get output buffer generated by the model named Meta-Llama-3-8B-Instruct-f16.gguf in the stream mode.
[2024-08-06T21:56:34Z INFO  llama-core] Get output buffer generated by the model named Meta-Llama-3-8B-Instruct-f16.gguf in the stream mode.
[2024-08-06T21:56:35Z INFO  llama-core] Get output buffer generated by the model named Meta-Llama-3-8B-Instruct-f16.gguf in the stream mode.
[...]

PR #335, opened 2024-12-29
issues #199: Display spinning/loading animation while an LLM response…

laonger:
add a "…" animation to indicate the message is still streaming:

https://github.com/user-attachments/assets/4944add2-421f-43cb-830d-3b15c05f8e3c

### DIFF
--- a/src/chat/chat_line_loading.rs
+++ b/src/chat/chat_line_loading.rs
@@ -92,6 +92,15 @@ live_design! {
             }
         }
     }
+
+    AnswerLoading = {{AnswerLoading}} {
+        width: 1,
+        height: 1,
+
+        flow: Down,
+        animator: {}
+    }
+    
 }
 
 #[derive(Live, LiveHook, Widget)]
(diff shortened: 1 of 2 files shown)


joulei (changes requested):
Hi @laonger , thanks for working on this improvement.
The mechanism you're using for updating the text is simple and works, but you're creating a new widget which seems unnecessary.
The `AnswerLoading` widget is not really a part of the UI, you're just using it for keeping a timer.

I suggest you rather move that timer to `ChatLine` instead and delete `AnswerLoading`.
There's also some renaming that would be helpful, leaving some comments.

src/chat/chat_line_loading.rs
```diff
@@ -167,3 +176,76 @@ impl ChatLineLoadingRef {
...
+                " ...".to_string()
+            },
+            3 => {
+                " ....".to_string()
+            },
+            4 | 5 | _ => {
+                "".to_string()
+            }
+        }
+    }
```
joulei, line 249: This doesn't update animation, and does not use animation as we understand it in Makepad. I would rename it to something like `get_animation_frame_text` which more clearly describes that this method returns a string

src/chat/chat_line_loading.rs
```diff
@@ -167,3 +176,76 @@ impl ChatLineLoadingRef {
...
     }
 }
+
+#[derive(Live, LiveHook, Widget)]
+pub struct AnswerLoading {
+    #[deref]
+    view: View,
+
+    #[rust]
+    couser: usize,
```
joulei, line 186: Took me a second to understand this field (I think there's also a typo here), I think we should rename it so something more clear like `animation_frame` or something that more clearly describes that this is used to count the amount of dots.

joulei:
Closing this due to inactivity (the codebase has changed tremendously), we might re-implement this in MolyKit in the future.

PR #336, opened 2024-12-30
Fix the response text parsing and encoding issue when the moly and mofa agents communicate locally.

ZhangHanDong:
The main fixes are for two issues:

1. Fix the response text parsing and encoding issue when the moly and mofa agents communicate locally.
2. If access to OpenAI or Claude is restricted in certain regions, a VPN is typically used, so communication with the local MoFA service here must bypass the proxy. 

image


joulei:
Hi @ZhangHanDong thanks for the PR. The proxy bypass looks good.
Also the changes or parsing seem to work but I'll wait before merging because this change is unexpected.
Until recently the response from MoFa servers (both locally and the remote servers) was formatted differently, we were following a different protocol. I'll ask the MoFa team about this.

PR #338, opened 2024-12-31
Fix the incompatibility issue of libwasmedge.0.dylib on macos x86_64

yangcancai:
## Fix the crash on macos x86_64
```shell
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               _moly_app [22638]
Path:                  /Applications/Moly.app/Contents/MacOS/_moly_app
Identifier:            _moly_app
Version:               ???
Code Type:             X86-64 (Native)
Parent Process:        Exited process [22637]
User ID:               501

Date/Time:             2024-12-31 11:29:13.7184 +0800
OS Version:            macOS 14.4.1 (23E224)
Report Version:        12
Bridge OS Version:     8.4 (21P4222)
Anonymous UUID:        38FD8E42-55B9-6552-B62A-10C3E31EE872

Sleep/Wake UUID:       A0EBF6FF-3DDF-4AB0-AF3A-BAF66E8CF9F7

Time Awake Since Boot: 990000 seconds
Time Since Wake:       11252 seconds

System Integrity Protection: disabled

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: @rpath/libwasmedge.0.dylib
Referenced from: <EE478A10-7904-3190-B4B7-50975A5D71B1> /Applications/Moly.app/Contents/MacOS/_moly_app
Reason: tried: '/Applications/Moly.app/Contents/Frameworks/libwasmedge.0.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64h' or 'x86_64')), '/Applications/Moly.app/Contents/Frameworks/libwasmedge.0.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64h' or 'x86_64'))
(terminated at launch; ignore backtrace)

Thread 0 Crashed:
0   dyld                          	    0x7ff8196c9b3e __abort_with_payload + 10
1   dyld                          	    0x7ff8196e2bb7 abort_with_payload_wrapper_internal + 82
2   dyld                          	    0x7ff8196e2be9 abort_with_payload + 9
3   dyld                          	    0x7ff81966f2e5 dyld4::halt(char const*, dyld4::StructuredError const*) + 335
4   dyld                          	    0x7ff81966c4e2 dyld4::prepare(dyld4::APIs&, dyld3::MachOAnalyzer const*) + 4099
5   dyld                          	    0x7ff81966b2ff start + 1839


Thread 0 crashed with X86 Thread State (64-bit):
  rax: 0x0000000002000209  rbx: 0x0000000000000001  rcx: 0x00007ff7bfbd9578  rdx: 0x00007ff7bfbd99e0
  rdi: 0x0000000000000006  rsi: 0x0000000000000001  rbp: 0x00007ff7bfbd95c0  rsp: 0x00007ff7bfbd9578
   r8: 0x00007ff7bfbd95e0   r9: 0x0000000000000000  r10: 0x000000000000005f  r11: 0x0000000000000246
  r12: 0x0000000000000000  r13: 0x00007ff7bfbd99e0  r14: 0x0000000000000006  r15: 0x000000000000005f
  rip: 0x00007ff8196c9b3e  rfl: 0x0000000000000246  cr2: 0x0000000000000000
  
Logical CPU:     0
Error Code:      0x02000209 
Trap Number:     133
```

### DIFF
--- a/packaging/before-packaging-command/src/main.rs
+++ b/packaging/before-packaging-command/src/main.rs
@@ -171,13 +171,21 @@ fn download_wasmedge_macos(version: &str) -> std::io::Result<()> {
     // Command 1: Create the destination directory.
     let dest_dir = std::env::current_dir()?.join("wasmedge");
     fs::create_dir_all(&dest_dir)?;
-
+    let mut wasmedge_download_url = String::new();
+    let mut wasi_nn_download_url = String::new();
+    if std::env::consts::ARCH == "x86_64" {
+        wasmedge_download_url = format!("https://github.com/WasmEdge/WasmEdge/releases/download/{}/WasmEdge-{}-darwin_x86_64.tar.gz", version, version);
+        wasi_nn_download_url = format!("https://github.com/WasmEdge/WasmEdge/releases/download/{}/WasmEdge-plugin-wasi_nn-ggml-{}-darwin_x86_64.tar.gz", version, version);
+    } else {
+        wasmedge_download_url = format!("https://github.com/WasmEdge/WasmEdge/releases/download/{version}/WasmEdge-{version}-darwin_arm64.tar.gz");
+        wasi_nn_download_url = format!("https://github.com/WasmEdge/WasmEdge/releases/download/{version}/WasmEdge-plugin-wasi_nn-ggml-{version}-darwin_arm64.tar.gz");
+    }
     // Command 2: Download and extract WasmEdge.
     {
         println!("Downloading wasmedge v{version} to: {}", dest_dir.display());
         let curl_script_cmd = Command::new("curl")
             .arg("-sSfL")
-            .arg("https://github.com/WasmEdge/WasmEdge/releases/download/0.14.0/WasmEdge-0.14.0-darwin_arm64.tar.gz")
+            .arg(wasmedge_download_url)
             .stdout(Stdio::piped())
             .spawn()?;
 
@@ -209,7 +217,7 @@ fn download_wasmedge_macos(version: &str) -> std::io::Result<()> {
         println!("Downloading wasmedge v{version} WASI-NN plugin to: {}", plugin_dest_dir.display());
         let curl_script_cmd = Command::new("curl")
             .arg("-sSfL")
-            .arg("https://github.com/WasmEdge/WasmEdge/releases/download/0.14.0/WasmEdge-plugin-wasi_nn-ggml-0.14.0-darwin_arm64.tar.gz")
+            .arg(wasi_nn_download_url)
             .stdout(Stdio::piped())
             .spawn()?;
 
--- a/packaging/before-packaging-command/src/main.rs
+++ b/packaging/before-packaging-command/src/main.rs
@@ -65,6 +65,7 @@ fn main() -> std::io::Result<()> {
     let mut is_before_packaging = false;
     let mut is_before_each_package = false;
     let mut host_os_opt: Option<String> = None;
+    let mut arch_opt: Option<String> = None;
 
     let mut args = std::env::args().peekable();
     while let Some(arg) = args.next() {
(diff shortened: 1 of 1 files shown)


L-jasmine:
LGTM

packaging/before-packaging-command/src/main.rs
```diff
@@ -171,13 +171,21 @@ fn download_wasmedge_macos(version: &str) -> std::io::Result<()> {
     // Command 1: Create the destination directory.
     let dest_dir = std::env::current_dir()?.join("wasmedge");
     fs::create_dir_all(&dest_dir)?;
-
+    let mut wasmedge_download_url = String::new();
+    let mut wasi_nn_download_url = String::new();
+    if std::env::consts::ARCH == "x86_64" {
```
kevinaboos, line 176: Don't we need to check the *target* architecture, not the architecture of the build machine? This assumes we never do any cross compilation and only build for the current arch.

packaging/before-packaging-command/src/main.rs
```diff
@@ -171,13 +171,21 @@ fn download_wasmedge_macos(version: &str) -> std::io::Result<()> {
     // Command 1: Create the destination directory.
     let dest_dir = std::env::current_dir()?.join("wasmedge");
     fs::create_dir_all(&dest_dir)?;
-
+    let mut wasmedge_download_url = String::new();
```
kevinaboos, line 174: the canonically Rusty way to do this is to either declare a variable without initializing it, or return a tuple from the following block.
You can just do:
```rust
let wasmedge_download_url: String;
let wasi_nn_download_url: String;
```

packaging/before-packaging-command/src/main.rs
kevinaboos: Or, at the very least, allow it to be overridden with a CLI arg or env var.

kevinaboos (changes requested):
Semantically it's fine, but I left a few small questions.

packaging/before-packaging-command/src/main.rs
yangcancai: Got it, will changed it the same with host_os.

kevinaboos:
Ok the code looks good now. Shouldn't we also document how to use this new `arch` argument, both in the code and in the Readme section about packaging? Otherwise it's unlikely that anyone will correctly use it.

Also, just to clarify, this seems to be only intended to support macOS x86_64 builds, right? If so, we should state that explicitly in the docs.

yangcancai:
`cargo packager --release` seems unable to pass arg. because `before-packaging-command` is hardcoded in Cargo.toml. Is it possible to use env instead?
For example, `ARCH=x86_64 cargo packager --release` passes arch, which can be obtained through `env::var("ARCH")` in the code.

kevinaboos:
Correct, you would not pass these arguments to the invocation of `cargo-packager` itself. You would add them into the invocation of the `before-packaging-command`, which is defined in the Cargo.toml file. 

But yes, and env var is fine too.

kevinaboos:
You shouldn't actually even need this to be a CLI arg or env var. Can't you just take it from the target spec? That is, if you're building for an aarch64 target, you'll already know that at the cargo and rust layer, based on the target triple selected at build time. So you can just use `cfg(target_arch)`, right?

yangcancai:
Nice, I agree with you, using cfg(target_arch) will be more rusty.

kevinaboos (approved):
Looks great! 

I haven't tested the code myself, so I won't merge it in, but will approve these changes.

PR #341, opened 2025-01-04
Add a confirmation dialog for MoFa server deletion

Guocork:


src/settings/mofa_settings.rs
```diff
@@ -2,6 +2,9 @@ use makepad_widgets::*;
 
 use crate::data::chats::{MofaServer, MofaServerConnectionStatus};
 use crate::data::store::Store;
+use crate::shared::modal::ModalWidgetExt;
```
Guocork, line 5: There's something I don't quite understand. When using the `modal` under `shared`, why is it necessary to include `ModalWidgetExt` for it to work properly?

joulei: I've wondered the same, let me double check and get you a proper answer

joulei (approved):
Good work @Guocork, looks good!

PR #357, opened 2025-01-24
Prompt example

Guocork:


src/chat/chat_panel.rs
```diff
@@ -418,13 +501,28 @@ impl Widget for ChatPanel {
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
         self.update_view(cx, scope);
 
+        // Read a file from JSON.
+        let file_content = fs::read_to_string("./resources/files/prompt_data.json").expect("Failed to read JSON file");
```
alanpoon, line 505: You should not be doing computation expensive operation here in the draw_walk as draw_walk runs multiple times.

Guocork: Thanks for your review! I noticed that, and I'll fix it in future commits.

src/chat/chat_panel.rs
```diff
@@ -446,6 +553,32 @@ impl WidgetMatchEvent for ChatPanel {
...
+
+        if prompt3.clicked(actions) {
+            self.prompt_example = prompt3.text();
+            self.prompt_input(id!(main_prompt_input)).set_text_and_redraw(cx, &self.prompt_example);
+        }
+
+        if prompt4.clicked(actions) {
+            self.prompt_example = prompt4.text();
+            self.prompt_input(id!(main_prompt_input)).set_text_and_redraw(cx, &self.prompt_example);
+        }
```
Guocork, line 580: Hello @joulei , When adding this feature, I ran into a problem. I couldn't capture the corresponding click event because the components in the panel are arranged using Overlay. I feel like it's an event propagation issue across different layers. I tested click events at various levels and found that only the events on `empty_conversation `couldn't be captured. I'm not sure why that's happening.

src/chat/chat_panel.rs
```diff
@@ -994,3 +1135,17 @@ fn get_chat_messages(store: &Store) -> Option<Ref<Vec<ChatMessage>>> {
 fn get_chat_id(store: &Store) -> Option<ChatID> {
     get_chat(store).map(|chat| chat.borrow().id)
 }
+
+ fn get_prompts() -> Vec<String> {
```
Guocork, line 1139: @joulei I originally read the prompt from JSON inside `draw_walk`, but since `draw_walk` runs in a loop, the data reading should only happen once when `empty_conversation` appears. So, I can't put that operation inside `draw_walk`. I tried extracting and wrapping the data reading into a separate function and tested a few approaches, but none of them worked as expected.

src/chat/chat_panel.rs
joulei: Ok I'll take a look at what's capturing the click events and preventing them from reaching `empty_conversation`, thanks!

PR #362, opened 2025-02-08
Add scroll bar  in PortalList

Guocork:


src/chat/chat_panel.rs
```diff
@@ -326,9 +326,7 @@ live_design! {
 
                 chat = <PortalList> {
                     margin: { bottom: 15 }
-                    scroll_bar: {
-                        bar_size: 0.0,
-                    }
+                    scroll_bar: <MolyScrollBar> {}
```
Guocork, line 329: Hello, @joulei  Right now, I've implemented the `scroll bar`, but there's an issue— in the chat panel, there's always a gap between the scroll bar and the bottom of the chat messages. I can drag it down with the mouse, but it snaps back to its original position. I tried fixing it, but no luck so far.

Guocork:
And also I want to make `Moly` scroll bar work like the one in Gmail, but right now, Makepad’s `ScrollBar` component doesn’t support setting a `visible` property. I’ve already opened an [issue](https://github.com/makepad/makepad/issues/655) for this in Makepad.

PR #366, opened 2025-02-17
remove unused code

Guocork:
This code is repetitive with the code above `handle_event`. I tested it and found that this code doesn't work. After I deleted it, the `selector` functionality didn't change.

### DIFF
--- a/src/chat/model_selector.rs
+++ b/src/chat/model_selector.rs
@@ -303,34 +303,6 @@ const MAX_OPTIONS_HEIGHT: f64 = 400.0;
 
 impl WidgetMatchEvent for ModelSelector {
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, _scope: &mut Scope) {
-        if let Some(fd) = self.view(id!(button)).finger_down(&actions) {
-            if fd.tap_count == 1 {
-                self.open = !self.open;
-
-                if self.open {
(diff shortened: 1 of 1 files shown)


src/chat/model_selector.rs
```diff
@@ -303,34 +303,6 @@ const MAX_OPTIONS_HEIGHT: f64 = 400.0;
 
 impl WidgetMatchEvent for ModelSelector {
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, _scope: &mut Scope) {
-        if let Some(fd) = self.view(id!(button)).finger_down(&actions) {
```
Guocork, line 306: It's redundant with the code [above.](https://github.com/moxin-org/moly/pull/366/files#diff-c0f61090fe55f67b3e07875b1c34cefab4cae9cda3434c1f85bfaeaeead9769bR212-R240) After testing, I found that this code doesn't correctly capture the event.

Guocork: This might be an issue with `Makepad`.

joulei: Thanks for the submission @Guocork , we ended up replacing a lot of code here and now the event handling works different in this case. Now the events are actually captured.

joulei: Yes, this was indeed an issue with event handling, that we have not solved since. thanks

PR #368, opened 2025-02-17
MolyKit: Re-usable AI widget kit

noxware:


joulei (approved):
While there are a few things that need fixing/reworking/extending for MolyKit to be used in Moly, we're merging into `dev` so that we can start integrating and fixing/improving (in parallel) the different issues.

As a side note, I think there are some approaches/mechanisms here that might be unexpected for Makepad beginners as they differ slightly from the "standard" Makepad approach. We can discuss this later, just keep in mind some APIs might change.

PR #376, opened 2025-03-05
Initial contextual capture

cass-uredium:
Remaining TODO:

- [x] Rework ownership of capture handler for shared use between URL handler and platform-specific.
- [ ] ~~Implement URL handler backend from existing partial demo.~~ (subsequent PR)
- [ ] ~~Clarify `CaptureHandler` drop semantics.~~
- [ ] ~~Populate error type.~~

### DIFF
--- a/src/chat/prompt_input.rs
+++ b/src/chat/prompt_input.rs
@@ -1,7 +1,10 @@
 use makepad_widgets::*;
 
 use crate::{
-    data::{chats::chat_entity::ChatEntityId, providers::RemoteModel, store::Store},
+    data::{
+        capture::CaptureAction, chats::chat_entity::ChatEntityId, providers::RemoteModel,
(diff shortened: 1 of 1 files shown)


cass-uredium:
will separate into a `moly-capture` crate and rebase w/ server modularization changes

cass-uredium:
Now with integration on the application side. Not sure if idiomatic Makepad, but otherwise ready for review. URL handling can be added in a subsequent PR, as part of a larger effort.

cass-uredium:
(macOS build error is likely spurious and just needs a re-run)

joulei (approved):
Looks good!
We'll soon replace the `prompt_input` you updated here, with widgets from `MolyKit`.
I'll make sure to port the prompt input over when merging that. Thanks!

PR #380, opened 2025-03-13
Tiny fixes + Chat hooks rework

noxware:
# Changes

- Require Rust 1.81
- Chat tasks are executed/applied immediately instead of going through makepad's actions.
  - This solves issues with tasks requiring to have always up to date state on the widgets and users hooking expecting the same.
  - The hooking system now uses callbacks that can be set to modify tasks just before they are executed, or to just get notified after they occurred.
    - You can still access the widget inside the hook with an `UiRunner` or communicating with raw makepad actions.

### DIFF
--- a/moly-kit/src/widgets/chat.rs
+++ b/moly-kit/src/widgets/chat.rs
@@ -1,7 +1,6 @@
 use futures::{stream::AbortHandle, StreamExt};
 use makepad_widgets::*;
 use std::cell::{Ref, RefMut};
-use std::sync::RwLock;
 use utils::asynchronous::spawn;
 
 use crate::utils::events::EventExt;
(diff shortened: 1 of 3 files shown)


moly-kit/src/widgets/chat.rs
```diff
@@ -357,23 +256,30 @@ impl Chat {
...
+                            message,
+                            messages.is_at_bottom(),
+                        )
+                    });
 
-                        // If the chunk contains citations, store them
-                        if let Some(cits) = &delta.citations {
-                            message.citations = cits.clone();
-                        }
+                    me.dispatch(cx, &mut ChatTask::UpdateMessage(index, message).into());
```
noxware, line 278: We can't dispatch `UpdateMessage` inside the closure cause the closure already borrows messages and with this changes, dispatch will executes almost immediately. 

TODO: Consider removing read_with and write_with, fix them, replace them, whatever.

moly-kit/src/widgets/chat.rs
```diff
@@ -408,54 +314,43 @@ impl Chat {
...
-            widget_uid: self.widget_uid(),
-        };
+    /// You can still hook into these tasks before they are executed if you set a hook with
+    /// [Chat::set_hook_before].
+    ///
+    /// Warning: Like other operation over makepad's [WidgetRef], this function may panic if you hold
+    /// borrows to widgets inside [Chat], for example [Messages] or [PromptInput]. Be aware when using
+    /// `read_with` and `write_with` methods.
+    // TODO: Mitigate interior mutability issues with many tricks or improving makepad.
+    pub fn dispatch(&mut self, cx: &mut Cx, tasks: &mut Vec<ChatTask>) {
```
noxware, line 324: The `&mut` on tasks is a pretty neat detail. This:
- Doesn't force you to create owned tasks (although most of the time you create them in place).
- Allows you to inspect the tasks after dispatching, seeing the effects of any hook that run.
- More importantly, this says "tasks MAY be modified", which differentiates this function from `perform`.

joulei (approved):

PR #382, opened 2025-03-14
MolyKit tiny cleanup

noxware:
- Address warnings at `messages.rs`
- Move citations code out of `messages.rs`
- Fix citations preview click event.
- Refactored a little `ChatDelta`.

### DIFF
--- a/moly-kit/src/clients/moly.rs
+++ b/moly-kit/src/clients/moly.rs
@@ -88,7 +88,7 @@ struct Choice {
 struct Completion {
     pub choices: Vec<Choice>,
     #[serde(default)]
-    pub citations: Option<Vec<String>>,
+    pub citations: Vec<String>,
 }
 
 #[derive(Clone, Debug, Default)]
@@ -194,7 +194,7 @@ impl BotClient for MolyClient {
         &mut self,
         bot: &BotId,
         messages: &[Message],
-    ) -> MolyStream<'static, Result<ChatDelta, ()>> {
+    ) -> MolyStream<'static, Result<MessageDelta, ()>> {
         let moly_messages: Vec<OutcomingMessage> = messages
             .iter()
             .filter_map(|m| m.clone().try_into().ok())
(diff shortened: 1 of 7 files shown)


moly-kit/src/clients/moly.rs
```diff
@@ -88,7 +88,7 @@ struct Choice {
 struct Completion {
     pub choices: Vec<Choice>,
     #[serde(default)]
-    pub citations: Option<Vec<String>>,
+    pub citations: Vec<String>,
```
noxware, line 91: @joulei was `Option` here due to some provider sending an explicit `null` here?

moly-kit/src/clients/moly.rs
```diff
@@ -268,17 +268,16 @@ impl BotClient for MolyClient {
...
 
-                    // Combine all partial choices content
-                    let content_delta = completion
+                    let body = completion
                         .choices
                         .iter()
                         .map(|c| c.delta.content.as_str())
                         .collect::<String>();
 
-                    let citations = completion.citations.clone();
```
noxware, line 278: nitpick: no need for clone since the struct will no longer be used after this

moly-kit/src/protocol.rs
```diff
@@ -80,11 +80,13 @@ pub struct Message {
...
-/// A new structure to hold both text delta and optional metadata like citations.
-#[derive(Clone, Debug)]
-pub struct ChatDelta {
-    pub content_delta: String,
-    pub citations: Option<Vec<String>>,
+/// A delta response for an existing [Message].
+#[derive(Clone, Debug, Default)]
+pub struct MessageDelta {
+    /// Delta for the [Message::body] field.
+    pub body: String,
```
noxware, line 87: `content_delta` sound like OpenAI (to which `protocol.rs` is not tied) and is not consitent with `Message`, that would be the justification for this change.

moly-kit/src/protocol.rs
```diff
@@ -80,11 +80,13 @@ pub struct Message {
...
+    /// Delta for the [Message::citations] field.
+    pub citations: Vec<String>,
```
noxware, line 89: Technically speaking, empty vecs are not heap allocated, so there is little difference on wrapping them with option, UNLESS there is a semantic meaning for this. 

Was this option controlling some behavior that I'm missing?

moly-kit/src/clients/moly.rs
joulei: Mhh no, I think this is leftover from an earlier iteration of this, or perhaps I was missing the `#[serde(default)]` and needed an option for the de-serialization to not break, this is fine to remove now.

joulei (approved):
Thanks for the cleanup!

PR #387, opened 2025-03-21
MolyKit Integration

joulei:
TODO

- [x] Add MolyKit Chat
- [x] Integrate Chat with ModelSelector
- [x] Integrate Chat with Providers dashboard
- [x] Integrate Chat with ChatHistory (wire with persistence, message updates, deletion, etc.)
- [x] Update MolyServer to be compatible with MolyKit OpenAI client (handle load/eject automatically) Depends on https://github.com/moxin-org/moly-server/pull/6
- [x] Fix styling
- [x] Make MolyServer a regular provider, instead of a separate entity coupled with files
- [x] Fix: whenever a new provider is added, its models show without a header in the selector
- [x] Fix broken methods in Moly's MolyClient (mainly download progress)
- [x] Fix chat with model buttons not working (my models and landing screen)
- [x] Remove/rework model loading
- [x] Add support for DeepInquire in MolyKit
- [x] Remove unused files - ChatPanel, ChatLine, etc.
- [x] Replace ChatEntityId and MessageChat with MolyKit abstractions (currently just mapping them)
- [x] MolyServer should handle model eject before file deletion https://github.com/moxin-org/moly-server/pull/7

Not a blocker for this PR:
- [x] Use system messages in Chat to report errors (main plumbing is  ready)
- [ ] Move MolyServer related features into it's own view
- [ ] Bug: The first chat in history shows as selected but is not loaded into the chat
- [ ] Some abstraction/API changes in MolyKit for more ergonomic and/or efficient use
- [ ] Override MolyKit's prompt input with Moly's (supports @model, we might not actually do this for a while)


moly-kit/src/clients/openai.rs
```diff
@@ -232,12 +234,12 @@ impl BotClient for OpenAIClient {
...
     fn send_stream(
         &mut self,
-        bot: &BotId,
+        bot: &Bot,
         messages: &[Message],
     ) -> MolyStream<'static, ClientResult<MessageDelta>> {
         let inner = self.0.read().unwrap().clone();
 
-        let url = format!("{}/v1/chat/completions", inner.url);
+        let url = format!("{}/chat/completions", inner.url);
```
joulei, line 242: There are OpenAI compatible providers that don't follow the same versioning scheme, so MolyKit shouldn't assume `/v1`

src/data/providers.rs
```diff
@@ -16,54 +15,48 @@ pub struct Provider {
...
-pub struct RemoteModel {
-    pub id: RemoteModelId,
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct ProviderBot {
+    pub id: BotId,
     pub name: String,
     pub description: String,
     pub provider_url: String,
     pub enabled: bool,
 }
```
joulei, line 39: If we want to add `enabled` flag to MolyKit's `Bot` (and actually use it) then we can get rid of this struct on the App side (at least for now we don't need any other extras)

src/data/store.rs
```diff
@@ -464,8 +281,16 @@ impl Store {
...
         self.chats.insert_or_update_provider(provider);
+        // Update in preferences (persist in disk)
         self.preferences.insert_or_update_provider(provider);
+        // Update in MolyKit (to update the API key used by the client, if needed)
+        if let Some(_bot_repo) = &self.bot_repo {
+            // Because MolyKit does not currently expose an API to update the clients, we'll remove and recreate the entire bot repo
+            // TODO(MolyKit): Find a better way to do this
+            self.bot_repo = None;
+        }
     }
```
joulei, line 294: I think `BotRepo` should be an actual repository-like interface and not a client interface, it might still hold a main
client/multi_client, but the crate user should be able to update the clients (add new ones, update existing ones, remove, etc.)
It would also be helpful if `BotRepo` can expose the bots without re-fetching every time. 

Or at least, I think we should rename `BotRepo` to something else as it does not follow a repository pattern. Even though I know it's just a client or client aggregator, I often found myself trying to use it like a repo.

src/settings/moly_server_settings.rs
```diff
@@ -1,11 +1,6 @@
 use makepad_code_editor::code_view::CodeViewWidgetExt;
 use makepad_widgets::*;
```
joulei, line 3: This file will actually change drastically once we move the MolyServer related features into it's own view. 
these changes are just placeholders, as now the connection settings are handled in the providers dashboard.

src/data/openai_client.rs
```diff
@@ -136,60 +112,8 @@ impl OpenAIClient {
     ///
     /// The loop continues until the command channel is closed or an unrecoverable error occurs.
     fn process_agent_commands(command_receiver: mpsc::Receiver<ProviderCommand>, address: String, api_key: Option<String>) {
-        let rt = tokio::runtime::Runtime::new().unwrap();
-        let mut current_request: Option<JoinHandle<()>> = None;
-
         while let Ok(command) = command_receiver.recv() {
             match command {
-                ProviderCommand::SendMessage(task, model, tx) => {
```
joulei, line 144: Since these clients are now only fetching models, their API (and the traits they're implementing, commands, etc) can be simplified a lot.

Ideally we would use MolyKit features to fetch for the models, there where some edge cases but I think now it might be replaceable. The main edge case (for why I didn't initially wire things like the providers dashboard with MolyKit `bots()`) is MoFa and DeepInquire not supporting /v1/models, so their clients hardcode the response, we should work around it and use MolyKit clients instead.
Since it's now a huge deal right now, I don't want to block this PR with this issue and kept these clients around, we can iterate over this once merged.

moly-kit/src/widgets/messages.rs
```diff
@@ -245,12 +254,19 @@ impl Messages {
...
                         item.message_loading(id!(text.loading)).animate(cx);
-
+                        item
+                    } else if message.has_stages() {
+                        // Use specialized DeepInquireBotLine for messages with stages
+                        let item = list.item(cx, index, live_id!(DeepInquireBotLine));
+                        
+                        // Update the DeepInquireBotLine with the message content
+                        item.as_deep_inquire_bot_line().set_message(cx, message, message.is_writing);
+
```
joulei, line 269: I didn't have much time for it now, but here's where I think we should have an API that allows MolyKit to defer/delegate the message rendering to a BotLine widget, that uses the base one by default, but app devs can extend.
So that you can create your own very custom client, that uses it's own formatting, and then Messages widget says "ok this client maps to this line renderer".
I know there's an option to make everything more dynamic through text (similar to thinking blocks) but I think there are great benefits to having a type safe implementation for providers/messages that are complex.
**tl;dr:** we should think about this, and make it in a way that DeepInquire serves as an example on how to add other very custom clients.

noxware (approved):

PR #392, opened 2025-03-26
Make moly pop up a notification box after a data stream interruption and retain the previous text in the UI. 

ZhangHanDong:
The issue where the UI does not retain rendered content during data interruptions has been fixed. 

Additionally, I resolved a parsing error in the latest dev code where the `id` could not be parsed, so I changed it to support optional serialization.

Screenshot 2025-03-26 at 17 45 54

Screenshot 2025-03-26 at 17 47 54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ preferences.json
 /wasmedge
 /WasmEdge*
 .DS_Store
+CLAUDE.md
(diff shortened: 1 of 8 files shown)


joulei:
@ZhangHanDong thanks for your PR! Unfortunately at the time you made these changes we were also replacing a lot other this functionality in Moly. We have now tracked these issues in MolyKit, and I used your network popup implementation to add some error handling for MolyServer connections. So while we won't merge this PR it was useful for tracking the issue fixes.
Thanks!

PR #409, opened 2025-04-10
Unhashed but composed bot id 

noxware:
# Changes
- Remove `hex` and `sha2` dependencies.
- `BotId`s are still created from a provider (discriminant) and an id from that provider.
  - But now it's not hashed improving debugging as you see these components when printed.
  - Now you can recover these sub components.
- `model_id` and `provider_id` in `Bot` are no longer necessary as they can be accessed like `bot.id.id()` and `bot.id.provider()` when necessary.

# Note about id encoding
In order to allow any characters inside the components of the id, while allowing these components to be rescued, the id is encoded without separator characters. Instead the size of the first component is encoded into the id itself allowing you to directly extract the slice from the string without handling escaping or allocating new strings.

# Note about id composition
Having `bot.id.id()` may seem weird at first but I can't think of other concise name. [LayerId](https://docs.rs/egui/latest/egui/layers/struct.LayerId.html) on egui looks similar.

### DIFF
--- a/src/data/chats/chat.rs
+++ b/src/data/chats/chat.rs
@@ -8,7 +8,6 @@ use crate::data::filesystem::{read_from_file, write_to_file};
 
 pub type ChatID = u128;
 
-
 #[derive(Debug, Default, Copy, Clone, Serialize, Deserialize)]
 enum TitleState {
     #[default]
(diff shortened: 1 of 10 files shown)


src/data/chats/chat.rs
```diff
@@ -103,8 +102,9 @@ impl Chat {
                 // Fallback to last_used_file_id if last_used_entity is None.
                 // Until this field is removed, we need to keep this logic.
                 let associated_bot = data.associated_bot.or_else(|| {
+                    // TODO: What's the provider here?
                     data.last_used_file_id
-                        .map(|file_id| BotId::from(file_id.as_str()))
+                        .map(|file_id| BotId::new(file_id.as_str(), ""))
```
noxware, line 107: @joulei

joulei (approved):

PR #423, opened 2025-04-21
Remove deep inquire specifics from protocol.rs

noxware:
# Main changes

- `protocol.rs` simplicity recovered and all deep inquire specifics removed from there.
  - Provider's specific content format can now be encoded into `MessageContent`'s `data` field.
    - This field is of type `String` due to limitations documented in the code. This should probably be optimized further.
  - This required a lot of cascade changes to purge deep inquire specifics from all moly, see first commit of the PR to see the "collateral damage".
- Clients do not stream deltas anymore. Instead they stream full messages content, so they can add, remove or replace content as they please. 

# Incomplete changes
- The ability to replace markdown content of a message with a custom widget provided by a client. 
  - For now, I preserved `DeepInquireBotLine` as a whole and its inclusion in `messages.rs`, but `DeepInquireContent` should be used instead once this feature is fully implemented.
  
# Other changes
- Simplified a bit the sse parsing but could be improved more.


joulei (approved):
Great job! Thanks for the cleanup
The only thing missing here is that I was previously showing the message loading animations even after the first stages arrived, and until we got a completion message.
I can add that back separately or we might even skip it if we can get the text streaming to work nicely enough.

PR #426, opened 2025-04-28
Move message content out

noxware:
# Changes

- Take out message content from messages.rs.
- Simplify related DSLs, like chat lines DSL.
- Reuse the message content widget inside deep inquire line for now.


noxware:
Due changes in deep inquire, this PR is closed favoring #452 which doesn't touch it that much.

PR #444, opened 2025-05-02
Fix provider modal radio buttons

noxware:
# Changes
- Radio button styling
  - Black text.
  - Border and background colors matching text input.
  - Use primary/accent color as marker.
- Added missing values for moly server and deep inquire.
- Auto format

### DIFF
--- a/src/settings/add_provider_modal.rs
+++ b/src/settings/add_provider_modal.rs
@@ -392,12 +392,19 @@ impl WidgetMatchEvent for AddProviderModal {
         }
 
         let selected = self
-            .radio_button_set(ids!(radios.radio_openai, radios.radio_mofa))
+            .radio_button_set(ids!(
+                radios.radio_openai,
+                radios.radio_mofa,
+                radios.radio_deepinquire,
+                radios.radio_moly_server
+            ))
             .selected(cx, actions);
         if let Some(selected) = selected {
             self.selected_provider = match selected {
                 0 => Some(ProviderType::OpenAI),
                 1 => Some(ProviderType::MoFa),
+                2 => Some(ProviderType::DeepInquire),
+                3 => Some(ProviderType::MolyServer),
                 _ => Some(ProviderType::OpenAI),
             };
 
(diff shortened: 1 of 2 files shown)


src/settings/add_provider_modal.rs
```diff
@@ -93,10 +96,45 @@ live_design! {
...
+            mark_color_hover: (TRANSPARENT)
+            mark_color_active: (PRIMARY_COLOR)
+            // This is the color of the tab focus, however there seem to be a bug
+            // cause if you select and hover-out, this color is used until you
+            // press tab. So for now let's just match the active color.
+            mark_color_focus: (PRIMARY_COLOR)
+            mark_color_active_focus: (PRIMARY_COLOR)
+            mark_color_active_hover: (PRIMARY_COLOR)
+        }
     }
```
noxware, line 138: this

src/settings/add_provider_modal.rs
```diff
@@ -361,18 +389,26 @@ impl WidgetMatchEvent for AddProviderModal {
...
+            .selected(cx, actions);
         if let Some(selected) = selected {
             self.selected_provider = match selected {
                 0 => Some(ProviderType::OpenAI),
                 1 => Some(ProviderType::MoFa),
+                2 => Some(ProviderType::DeepInquire),
+                3 => Some(ProviderType::MolyServer),
                 _ => Some(ProviderType::OpenAI),
             };
         }
```
noxware, line 408: this

joulei (approved):

src/shared/styles.rs
```diff
@@ -5,11 +5,14 @@ live_design! {
     use link::widgets::*;
     use makepad_draw::shader::std::*;
 
+    pub TRANSPARENT = #0000
+    pub PRIMARY_COLOR = #3A7CA5
```
noxware, line 9: Added these ones for convenience. 

`#0000` are not very legible if they are surrounded bt `#000`. 

Defining a single primary color without accents or secondaries is enough. In a minimalist design you can use this color for everything that is "default" or "active", like focused text input borders (active), call to action button (default), etc and then just use blank with borders for everything else :P

PR #464, opened 2025-05-13
Improve server usage documentation

cass-uredium:
Provide more direct instructions for server usage and clean up surrounding language.

There is still room for improvement here, but this should suffice relative to its alpha state. In the future, something like an installer, Docker image, etc., would be more user-friendly.


joulei (approved):
Thanks!

PR #474, opened 2025-05-14
Web

noxware:
Quick changes to make moly compile and run on the web. Many 1:1 translations without to much thinking.

Still has stuff that doesn't work.

Use `cargo makepad wasm --bindgen run -p moly` to run.


Cargo.toml
```diff
@@ -17,36 +17,37 @@ description = "Desktop app for downloading and chatting with AI LLMs"
 
 ## Rename the binary to `_moly_app` to avoid naming conflicts
 ## with the `moly` binary defined by the `moly-runner` crate.
-[[bin]]
-name = "_moly_app"
-path = "src/main.rs"
```
noxware, line 22: does not work with `cargo makepad wasm` but preserved unless moly runner disappears

src/data/downloads/download.rs
```diff
@@ -47,66 +46,54 @@ impl Download {
...
+                                    file_id: file_id.clone(),
+                                    kind: DownloadFileActionKind::Error,
+                                });
+                                eprintln!("Error downloading file: {:?}", err);
+                                break;
                             }
-                            // Don't break immediately - let any pending messages be processed
-                            // Just drop the handle and continue the loop
-                            progress_handle = tokio::spawn(async {});
-                        }
```
noxware, line 109: idk the exact way this is expected to work but seems to work with the new changes

src/data/filesystem/web.rs
```diff
@@ -0,0 +1,53 @@
+use anyhow::{anyhow, Result};
+use std::{
+    collections::HashMap,
+    path::PathBuf,
+    sync::{LazyLock, RwLock},
+};
+
+static FS: LazyLock<RwLock<HashMap<PathBuf, String>>> =
+    LazyLock::new(|| RwLock::new(HashMap::new()));
```
noxware, line 9: As the same file may be represented with different `Path`s, and on web there is no way to canonicalize them, this is a very fragile and not nice solution, but is quick to implement and seems to be working for now.

It may be better to adapt callers of these functions to not try to use fs apis on the first place and instead use something more db-like (like a simple key-value storage or something).

src/data/moly_client.rs
```diff
@@ -73,30 +76,34 @@ impl MolyClient {
...
+        spawn(async move {
             let resp = client.get(&url).send().await;
             match resp {
                 Ok(r) => {
                     if r.status().is_success() {
                         match r.json::<Vec<Model>>().await {
                             Ok(models) => {
-                                let _ = tx.send(Ok(models));
+                                let _ = tx.unbounded_send(Ok(models));
                             }
```
noxware, line 92: just quick translations but probably `spawn` can be removed, the channel generic can be set as the result of the function, and the function it self could be marked as `async`.

may correct them later

src/data/search/faked_models.rs
noxware, line 1: I think we are not relying on this file too much and makes porting the app to the web harder as it tries to randomize the fake models.

src/data/store.rs
```diff
@@ -204,14 +214,42 @@ impl Store {
...
 
-    pub fn delete_file(&mut self, file_id: FileID) -> Result<()> {
-        self.chats.eject_model().expect("Failed to eject model");
-
-        self.downloads.delete_file(file_id.clone())?;
-        self.search
-            .update_downloaded_file_in_search_results(&file_id, false);
-
-        Ok(())
+    pub fn delete_file(&mut self, file_id: FileID) {
```
noxware, line 217: This kind of store functions that do something on the remote (async) but try to return data (sync) have it return value removed.

Ideally, the store should separate mutations from data reading. Going into async and try to return something sync makes little sense.

PR #490, opened 2025-05-28
Web fs

noxware:
## Changes
- Filesystem abstraction capable of working on native and web. 
  - Works over "adapters" that can be implemented with the `Adapter` trait.
    - On native uses `async_fs` crate.
    - On web uses `web_fs`, whose api is modeled as the `async_fs` one, and works on top of [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system).
- Store loading is now async, so moly will not show until loaded.
  - Note: On the long term, I'm worried by moly loading all chats on start.
  
## Notes

Native filesystem adapter could use `tokio` specific APIs while on web the OPFS could be used directly with some hassle through `web_sys`, but `async_fs` being agnostic of tokio, and `web_rs` matching its api was pretty convincing and productive. Can be changed if needed as adapters are generic.

## Future

After much research trying to preserve the way Moly works directly over files, I believe we will need a proper database with at least migrations support in the future to avoid breaking users data on app upgrades. This is specially important now since OPFS is not easy to modify from outside.

### DIFF
--- a/src/data/chats/mod.rs
+++ b/src/data/chats/mod.rs
@@ -7,7 +7,6 @@ use moly_protocol::data::*;
 use std::collections::HashMap;
 use std::{cell::RefCell, path::PathBuf};
 
-use crate::app::app_runner;
 use crate::shared::utils::filesystem;
 
 use super::moly_client::MolyClient;


src/shared/utils/filesystem/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+/// It's abstract but async natiure allows it to be used on restrictive environments
+/// like the web.
+///
+/// Writes (inside this instance) are queued and executed in order in favor of a
+/// more deterministic behavior.
+#[derive(Debug)]
+pub struct FileSystem<A: Adapter> {
+    adapter: Arc<futures::lock::Mutex<A>>,
+    write_queue: mpsc::Sender<(PathBuf, Vec<u8>, oneshot::Sender<Result<()>>)>,
+}
```
noxware, line 28: Notice how writes are ordered. Better be safe than sorry.

src/shared/utils/filesystem/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /// Get a list of the entry names in the given directory.
+    pub async fn list(&self, path: &Path) -> Result<Vec<String>> {
+        let mut adapter = self.adapter.lock().await;
+        adapter.list(path).await
+    }
+
+    /// Write some bytes content to a given path, creating any necessary directories.
+    // TODO: Is adapter responsability to create directories, but it shouldn't.
```
noxware, line 122: Yeah, maybe after merging

src/shared/utils/filesystem/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+            use adapters::web::WebAdapter;
+            static FS: LazyLock<FileSystem<WebAdapter>> = LazyLock::new(|| FileSystem::new(WebAdapter::default()));
+        } else {
+            use adapters::native::NativeAdapter;
+            static FS: LazyLock<FileSystem<NativeAdapter>> = LazyLock::new(|| FileSystem::new(NativeAdapter::default()));
+        }
+    }
+
+    FS.clone()
+}
```
noxware, line 161: So, this is what really matters of this PR. calling `filesystem::global()` anywhere in Moly will give you a singleton to the abstract filesystem implemented in this file.

src/shared/utils/filesystem/adapter.rs
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+        pub trait PlatformSpecifics: Send {}
+        impl<F, O> PlatformSpecifics for F
+        where
+            F: Future<Output = O> + Send,
+            O: Send,
+        {
+        }
+    }
+}
```
noxware, line 24: Since all ours futures will eventually end at moly kit `spawn` function, which requires `Send` depending on the platform due to `JsValue`s not being `Send` on the web but having multithreaded runtime on native which requires `Send`, this trick helps.

On my free time, I could verify that `reqwest` works well over tokio single threaded runtime so we could see if we can simplify the `spawn` function to not require `Send` on native (limiting it to the most restrictive platform which is the web)

PR #491, opened 2025-05-29
Small profile + web build instructions

noxware:


README.md
```diff
@@ -88,6 +88,30 @@ cd moly
 cargo run --release
 ```
 
+## Building and Running (web)
```
noxware, line 91: instructions written ignoring the current problem of chinese fonts not being copied into build directory and causing the app to not load

PR #524, opened 2025-06-25
Image generation

noxware:
# Changes
- Adds support for OpenAI image generation models to Moly Kit.
  - Implements the `BotClient` trait.
  - Uses attachments feature backbone, so improvements to the attachments feature (like previews and persistence) will improve this feature as well.
  - Can be tested through moly-mini.
- Adds OpenAI image generation provider settings main moly app.
  - Done, BUT **see PR comments for details**.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2034,6 +2034,7 @@ dependencies = [
  "chrono",
  "env_logger",
  "futures",
+ "js-sys",
  "log",
  "makepad-code-editor",
  "makepad-widgets",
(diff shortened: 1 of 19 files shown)


noxware:
The main moly app seems to identify providers by the URL causing collision between the OpenAI provider and the OpenAIImage one.

When I say "provider" I mean the ones in the main moly app, not the "clients" in moly kit. 

Idk the implications of changing how providers identify, and there are some lines that are not easy to adapt. So I want to keep this out of the scope of this PR, as we also have other problems with providers in moly that we may want to solve together.

For now, the url for image generation is prefixed with `#` and the provider/client will remove it when appropriate.

PR #538, opened 2025-08-02
Tracking PR for Dora backend support

haixuanTao:
This is a tracking PR of the changes I made to make dora backend work.

I'm not exactly sure how you want those changes to be integrated and we can make some nicer cherry-picked PR, but hope this gives a general overview.

FYI, I had to change some flags to make sure to not record the silence before the AI speaks to get a lower first answer latency.

### DIFF
--- a/moly-kit/src/clients/openai_realtime.rs
+++ b/moly-kit/src/clients/openai_realtime.rs
@@ -40,6 +40,7 @@ pub enum OpenAIRealtimeMessage {
 pub struct SessionConfig {
     pub modalities: Vec<String>,
     pub instructions: String,
+    pub model: String,
     pub voice: String,
     pub input_audio_format: String,
     pub output_audio_format: String,
@@ -351,6 +352,7 @@ impl OpenAIRealtimeClient {
                             RealtimeCommand::UpdateSessionConfig {
                                 voice,
                                 transcription_model,
+                                llm,
                             } => {
                                 log::debug!(
                                     "Updating session config with voice: {}, transcription: {}",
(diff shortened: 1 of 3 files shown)


joulei:
Thanks! this is very helpful.

I've ported most of these changes to https://github.com/moxin-org/moly/pull/539 which I've now merged
(with some differences, like the user chooses the LLM first, and dora realtime is aded as a new default provider)

https://github.com/user-attachments/assets/228ab937-6d77-4587-b3e1-0b9313e5be55

However I haven't managed to run this in my machine fully, I'm getting some errors:

image

perhaps something missing in my setup?

```
dora logs 019876a0-b5b3-784d-839a-6a39cc1e28a5 tts

/python3.11/site-packages/torch/nn/modules/rnn.py:123: UserWarning: dropout option adds dropout after all but last recurrent layer, so non-zero dropout expects num_layers greater than 1, but got dropout=0.2 and num_layers=1
  warnings.warn(
/python3.11/site-packages/torch/nn/utils/weight_norm.py:143: FutureWarning: `torch.nn.utils.weight_norm` is deprecated in favor of `torch.nn.utils.parametrizations.weight_norm`.
  WeightNorm.apply(module, name, dim)
Configuration file could not be loaded.
File contains no section headers.
file: '/Users/devlian/.config/pip/pip.conf', line: 1
'break-system-packages = true\n'
WARNING: Defaulting repo_id to hexgrad/Kokoro-82M. Pass repo_id='hexgrad/Kokoro-82M' to suppress this warning.
```

PR #539, opened 2025-08-04
Add a default Dora Realtime provider

yulnr:


haixuanTao:
I think if you:

```uv pip install https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.8.0/en_core_web_sm-3.8.0-py3-none-any.whl``` 

The problem should go away. I'm going to add it. Sorry!

haixuanTao:
See: https://github.com/hexgrad/kokoro/issues/160

PR #568, opened 2025-09-26
Remove renaming

haixuanTao:
Remove naming conflict in for wasm


joulei:
Renaming removed in #574

PR #626, opened 2026-01-15
feat: Convert long pasted text to attachments

### LINKED ISSUE #497: Send/display long texts as text attachments
If the user pastes a very long text on the chat screen, we want to imitate Claude's app where it attaches the pasted text as a file/attachment rather than a big blob of text taking too much screen space. We should set a reasonable limit that triggers this behavior, and make clear in the UI that text is being sent as an attachment.

yulnr:
Automatically converts text longer than 2000 characters into a text attachment to prevent large text blobs from taking up too much screen space.

This matches the behavior of Claude's app and provides a better UX when users paste long documents or code snippets.

Fixes #497

----

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/moly-kit/src/widgets/attachment_view.rs
+++ b/moly-kit/src/widgets/attachment_view.rs
@@ -42,12 +42,29 @@ live_design! {
         }
 
         text_preview_wrapper = <View> {
+            width: Fill,
+            height: Fill,
             padding: 8,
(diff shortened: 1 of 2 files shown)


joulei:
@claude this works fine, but is possible to make it so that it gets set as attachment on paste, rather than on sent?

